PR #3, opened 2020-06-08
Initial review PR for convenience of commenting

paultyng:


tfexec/errors.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+var usageRegexp = `Too many command line arguments|^Usage: .*Options:.*|Error: Invalid -\d+ option`
+
+func (e *ErrCLIUsage) Error() string {
+	return e.stderr
+}
+
+// catchall error
+type Err struct {
```
paultyng, line 68: We should drop this catchall and just raise the lower errors.

tfexec/main.go
```diff
@@ -0,0 +1,27 @@
...
+)
+
+// FindTerraform attempts to find a Terraform CLI executable.
+//
+// As a first preference it will look for the environment variable
+// TFEXEC_TERRAFORM_PATH and return its value. If that variable is not set, it will
+// look in PATH for a program named "terraform",
+// and, if one is found, return its absolute path.
+func FindTerraform() (string, error) {
+	if p := os.Getenv("TFEXEC_TERRAFORM_PATH"); p != "" {
```
paultyng, line 16: I don't know if this package should read environment, people can always pass it in.

paultyng: We may just want to restructure the binary finder/install to a strategy pattern or something, ie. look path, explicit path, env vars, install from releases, are all just slightly different strategies for this and may be combined, etc.

tfexec/options.go
```diff
@@ -0,0 +1,218 @@
+package tfexec
+
```
paultyng, line 2: docs on these should probably indicate what cli flag they represent

tfexec/terraform.go
```diff
@@ -0,0 +1,606 @@
...
+	workingDir  string
+	execVersion string
+	env         []string
+	logger      *log.Logger
+}
+
+// NewTerraform returns a Terraform struct with default values for all fields.
+// If a blank execPath is supplied, NewTerraform will attempt to locate an
+// appropriate binary on the system PATH.
+func NewTerraform(workingDir string, execPath string) (*Terraform, error) {
```
paultyng, line 27: this can just be `New`

tfexec/errors.go
```diff
@@ -0,0 +1,74 @@
...
+	case regexp.MustCompile(`Error: Could not satisfy plugin requirements`).MatchString(stderr):
+		return &ErrNoInit{stderr: stderr}
+	case regexp.MustCompile(`Error: No configuration files`).MatchString(stderr):
+		return &ErrNoConfig{stderr: stderr}
+	default:
+		return &Err{stderr: stderr}
+	}
+}
+
+type ErrNoSuitableBinary struct {
```
paultyng, line 23: Errors should probably be suffixed with `...Error` to mimic other Go conventions:

* https://pkg.go.dev/os?tab=doc
* https://pkg.go.dev/net/http?tab=doc#ProtocolError

etc...

tfexec/terraform.go
```diff
@@ -0,0 +1,606 @@
...
+	return &tf, nil
+}
+
+func (tf *Terraform) SetEnv(env map[string]string) {
+	var tfenv []string
+
+	// always propagate CHECKPOINT_DISABLE env var unless it is
+	// explicitly overridden with tf.SetEnv
+	if _, ok := env["CHECKPOINT_DISABLE"]; !ok {
+		env["CHECKPOINT_DISABLE"] = os.Getenv("CHECKPOINT_DISABLE")
```
paultyng, line 67: This should probably be done at execution time to ensure its still picked up even if environments aren't set?

tfexec/terraform.go
```diff
@@ -0,0 +1,606 @@
...
+	err := versionCmd.Run()
+	if err != nil {
+		fmt.Println(errBuf.String())
+		return "", fmt.Errorf("%s, %s", err, errBuf.String())
+	}
+
+	return outBuf.String(), nil
+}
+
+type initConfig struct {
```
paultyng, line 97: You may want to split out in to files per command.

tfexec/terraform.go
```diff
@@ -0,0 +1,606 @@
...
+	"log"
+	"os"
+	"strings"
+
+	tfjson "github.com/hashicorp/terraform-json"
+)
+
+type Terraform struct {
+	execPath    string
+	workingDir  string
```
paultyng, line 18: this should probably be referred to as the `rootModulePath` or something, its the working dir for TF execution, but only because the working dir is assumed to be the root module, so may be a clearer name.

tfexec/terraform.go
paultyng: As above, can probably rename `workingDir` to `rootModulePath` or something to be more explicit?

radeksimko: I think in this case we don't know if the path is actually going to even have any Terraform code? So should we be making assumption it's a root module? `terraform fmt` for example can run from anywhere, when the file content is piped into STDIN.

tfexec/terraform.go
radeksimko: I think this will largely depends on the command being executed. Most commands (`apply`, `plan`, `destroy`) do expect the workdir to be also root module, but some do not, so I'd be careful about early (from perspective of the workflow) assumptions here.

tfexec/terraform.go
```diff
@@ -0,0 +1,606 @@
...
+
+func (opt *VarOption) configureApply(conf *applyConfig) {
+	conf.vars = append(conf.vars, opt.assignment)
+}
+
+func (opt *DirOrPlanOption) configureApply(conf *applyConfig) {
+	conf.dirOrPlan = opt.path
+}
+
+func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
```
paultyng, line 258: This should return the state types

tfexec/terraform.go
```diff
@@ -0,0 +1,606 @@
...
+
+func (opt *LockOption) configurePlan(conf *planConfig) {
+	conf.lock = opt.lock
+}
+
+func (opt *DestroyFlagOption) configurePlan(conf *planConfig) {
+	conf.destroy = opt.destroy
+}
+
+func (tf *Terraform) Plan(ctx context.Context, opts ...PlanOption) error {
```
paultyng, line 419: This should return the plan types from terraform-json

tfexec/terraform.go
paultyng: Agree with Radek here, can disregard my comment.

tfexec/terraform.go
paultyng: Nevermind, this has an upstream dependency as discussed.

tfexec/terraform.go
paultyng: Nevermind, this has an upstream dependency as discussed.

tfexec/terraform.go
paultyng: Agree with Radek about the root module thing, working dir makes sense, but I still think this should just be `tfexec.New`.

paultyng:
Going to close this, I think we have checked off the major conversation items.

PR #7, opened 2020-06-15
Adds version output parsing

paultyng:


tfexec/version_test.go
```diff
@@ -0,0 +1,114 @@
...
+		},
+		{
+			mustVer("0.12.18"), map[string]*version.Version{
+				"null": mustVer("2.1.2"),
+			}, `
+Terraform v0.12.18
++ provider.null v2.1.2
+
+Your version of Terraform is out of date! The latest version
+is 0.12.26. You can update by downloading from https://www.terraform.io/downloads.html
```
radeksimko, line 89: I think the note could be suppressed by disabling checkpoint when running under test, which would make the tests less fragile.

radeksimko: And perhaps we should even disable checkpoint for `version` commands by default as the advice is ignored and it's not shown to the end-user anyway?

tfexec/version.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+var (
+	simpleVersionRe = `v?(?P<version>[0-9]+(?:\.[0-9]+)*(?:-[A-Za-z0-9\.]+)?)`
+
+	versionOutputRe         = regexp.MustCompile(`^Terraform ` + simpleVersionRe)
+	providerVersionOutputRe = regexp.MustCompile(`(\n\+ provider[\. ](?P<name>\S+) ` + simpleVersionRe + `)`)
+)
+
+// TODO: maybe add JSON output to the version command to simplify all of this?
```
kmoe, line 48: This is done in https://github.com/hashicorp/terraform/pull/25252 and will land in 0.13. https://github.com/hashicorp/terraform-exec/issues/13 tracks use of this feature in tfexec.

```suggestion
```

tfexec/version.go
```diff
@@ -0,0 +1,78 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/hashicorp/go-version"
+)
+
+// Version returns structured output from the `terraform version` command including both the Terraform CLI version
+// and any initialized provider versions.
+func (tf *Terraform) Version(ctx context.Context) (tfVersion *version.Version, providerVersions map[string]*version.Version, err error) {
+	// TODO: support re-invocation of this?
```
kmoe, line 16: I think `Version()` should always invoke `terraform version` - `tfexec` is intended as a way to invoke `terraform` CLI commands as transparently as possible.

If the consumer wants the cached versions, they can read the fields from the `Terraform` struct.

radeksimko: I agree with Katy here, I think that any caching responsibility should be up to the consumer, even if it's consumer within `terraform-exec` itself. e.g. I assume at some point we may need to cache and/or transparently expose the version here https://github.com/hashicorp/terraform-exec/blob/22dfd359c0f8d72deaecb08a0294851c4f622174/tfinstall/tfinstall.go#L133-L136

but any such caching/exposure decision should be made within that package, keeping `tfexec` itself simple and transparent.

paultyng: Yeah I agree (about re-invocation at least, I dislike a field for this, will put a new commit together with thoughts).

Especially now that this is more info than just the tf version (provider versions), re-invocation is much more useful.

paultyng:
This is based on #16

paultyng:
@kmoe rebased and added runtime tests for 0.11, 0.12, and 0.13.

paultyng:
The wintest failure is Go 1.12 it seems.

kmoe:
Yes, the Windows executor is using Go 1.12. It seems it's not happy with the use of `%w`.

paultyng:
All 🟢!

PR #8, opened 2020-06-23
tfinstall package

kmoe:
Adds `tfinstall.Find(...*findOption) (string, error)`, which locates or downloads a `terraform` binary according to the strategies provided. These are:

#### `ExactPath(execPath string)`  
Path to an existing `terraform` binary on the local filesystem.
#### `LookPath()` 
Attempt to locate a `terraform` binary on the local filesystem using `exec.LookPath()`.
#### `LatestVersion(installDir string)` 
Attempt to download the latest version of `terraform` and save it to the provided `installDir` (if blank, a temporary directory will be created).
#### `ExactVersion(tfVersion string, installDir string)` 
As above, but will only download the exact provided version of `terraform`.

## Example

`tfPath, err := tfinstall.Find(tfinstall.LookPath(), tfinstall.LatestVersion(""))` 

First, `tfinstall` will attempt to locate `terraform` on the local system path. If found, this path will be returned as `tfPath`. 

If the executable is not found on the local system path, `tfinstall` will download the latest version of `terraform` and return the path to the downloaded executable as `tfPath`.

An error is only returned if all provided strategies fail.


kmoe:
WIP, reconsidering API

tfinstall/tfinstall.go
```diff
@@ -0,0 +1,346 @@
...
+	var errors []string
+
+	// go through the options in order
+	// until a valid terraform executable is found
+	for _, opt := range opts {
+		p, err := opt.find()
+		if err != nil {
+			// TODO KEM make logging configurable
+			findErr := fmt.Sprintf("did not locate terraform executable using strategy %s (error: %s)", opt.name, err)
+			errors = append(errors, findErr)
```
radeksimko, line 137: Would it make sense to use something like go-multierror here?

tfinstall/tfinstall.go
```diff
@@ -0,0 +1,346 @@
...
+0G8fSbjYsIF+skY+zhNbY1MenfIPctB9X5iyW291mWW7rhhZyuqqxN2xnmPPgFmi
+QGd+8KVodadHABEBAAGJATwEGAECACYCGwwWIQSRpuf4XQXGVjC+8YlRhS2HNI/8
+TAUCXn0BRAUJEvOKdwAKCRBRhS2HNI/8TEzUB/9pEHVwtTxL8+VRq559Q0tPOIOb
+h3b+GroZRQGq/tcQDVbYOO6cyRMR9IohVJk0b9wnnUHoZpoA4H79UUfIB4sZngma
+enL/9magP1uAHxPxEa5i/yYqR0MYfz4+PGdvqyj91NrkZm3WIpwzqW/KZp8YnD77
+VzGVodT8xqAoHW+bHiza9Jmm9Rkf5/0i0JY7GXoJgk4QBG/Fcp0OR5NUWxN3PEM0
+dpeiU4GI5wOz5RAIOvSv7u1h0ZxMnJG4B4MKniIAr4yD7WYYZh/VxEPeiS/E1CVx
+qHV5VVCoEIoYVHIuFIyFu1lIcei53VD6V690rmn0bp4A5hs+kErhThvkok3c
+=+mCN
+-----END PGP PUBLIC KEY BLOCK-----`
```
radeksimko, line 56: Nitpick: It would be nice to perhaps move this key into a separate (go) file, just to reduce LOC here.

tfinstall/tfinstall.go
```diff
@@ -0,0 +1,346 @@
...
+		return err
+	}
+	_, err = openpgp.CheckDetachedSignature(el, data, sig)
+
+	return err
+}
+
+// verifySums verifies the hash of the downloaded file against the appropriate line
+// in SHA256SUMS
+func verifySums(terraformPath, sumsPath string) error {
```
radeksimko, line 289: It seems that we don't perform the checksum verification, i.e. this function is not being called from anywhere?

tfinstall/tfinstall.go
```diff
@@ -0,0 +1,346 @@
...
+	for {
+		line, err := rd.ReadString('\n')
+		if err != nil {
+			if err != io.EOF {
+				return err
+			}
+			break
+		}
+		parts := strings.Fields(line)
+		if parts[0] == "terraform" {
```
radeksimko, line 317: I may be misunderstanding the intentions here, but isn't the first part of checksum file the hash? e.g.

```
893050bcfc5e7445acd3a30f1500227b989b29cbd958ca64a8233589194a198d  terraform_0.12.28_darwin_amd64.zip
7f443c80d4793d3181cc579ef4c37a7cf76b63b316670b1abd7598ebe421dd44  terraform_0.12.28_freebsd_386.zip
9d3381adc015cda84687ccc4aed7714645857f901c301d7f5d20ecc78844f3fd  terraform_0.12.28_freebsd_amd64.zip
c8c76a215c7f088a547fdb318ff08a48ad45a6995a4b5b36cfcf05d01ed7c86d  terraform_0.12.28_freebsd_arm.zip
1c489282d86b16f2d5f89f38071c6dabd948a4ca7cc4e42915e604b82564f3a6  terraform_0.12.28_linux_386.zip
be99da1439a60942b8d23f63eba1ea05ff42160744116e84f46fc24f1a8011b6  terraform_0.12.28_linux_amd64.zip
ee1835429b681460e0adcc6f873a1d1390264647bae77151ba8211610dc86d83  terraform_0.12.28_linux_arm.zip
c5c32dc1896188a17f93e0e7aa3f160fe842ad0eb4e4761345e728b1ac26e8e2  terraform_0.12.28_openbsd_386.zip
d1818c9d5414b8fa5ed9618bf00abd605b016be7812e5b71661337693a47d165  terraform_0.12.28_openbsd_amd64.zip
a904775f596895ff84dcc76ffff68bebf0c115b39e21f9a612431fcadbc814da  terraform_0.12.28_solaris_amd64.zip
a6c8c2f4d20c38f3467cbc65e3d34d446751a675f4b1bfdf6be155baf14f8963  terraform_0.12.28_windows_386.zip
cd524b5b6b7cd9eec9c4e49aa37cbcb34ed1395876c212f6dde84c6e57d6ce1c  terraform_0.12.28_windows_amd64.zip

```

tfinstall/tfinstall.go
paultyng: Why don't we just stop at the first error, can the methods just return empty string and nil if not found?

tfinstall/tfinstall.go
kmoe: Oops, I wrote this before discovering the go-getter feature that does this automatically. Will remove this func.

kmoe:
Thanks for the `tfdisco` name, @radeksimko . I love it. I think it's a great name for the package overall, and the consumer can decide which parts of it to use. The language server need only run `tfdisco.Find(tfdisco.LookPath())` (or perhaps `tfdisco.Find(tfdisco.ExactPath("/special/path"), tfdisco.LookPath())` if you want to allow users to force the LS to prefer a particular binary by placing it at `"/special/path"`). 

I've simplified the error handling logic along the lines suggested by @paultyng. Errors encountered during each strategy will now cause `Find()` to error overall, except in the following two expected cases:

1. `tfdisco.LookPath()`: if a `terraform` executable is not found, fall through to the next strategy. (Of course if `LookPath()` is the final strategy, this means `Find()` will error overall, as it cannot return a valid path to an executable).

2. `tfdisco.ExactPath(path)`: We now stat the provided path, and fall through without error if it does not exist. This is based on how I imagine library consumers might use `ExactPath()` - to provide one or two default paths up front, in case Terraform is already installed in a well-known location.

radeksimko:
Sorry for the misunderstanding, but I was rather suggesting decoupling of "discovery" functions (i.e. `LookPath` and `ExactPath`) entirely to the `tfdisco` package and leaving the installation functions (`LatestVersion`, `ExactVersion`) in `tfinstall`.

I understand that the test framework needs the functionality "find or install", but I think that decision logic (whether to install) should either stay in the test framework or be decoupled to yet another layer if it really stays here.

e.g. I may want to do discovery earlier in the workflow, but install later, or something like that and the current coupling would make me do the discovery again, which seems unnecessary?

kmoe:
> e.g. I may want to do discovery earlier in the workflow, but install later, or something like that and the current coupling would make me do the discovery again, which seems unnecessary?

I don't quite understand your use case, and I'm reluctant to split into two packages here because I see this package as a natural grouping of `Find()` and its `findOption`s. Could you describe a bit more the LS workflow, and why it wouldn't be sufficient to do `tfdisco.Find(tfdisco.LookPath())` earlier, and `tfdisco.Find(tfdisco.LatestVersion())` (for instance) later?

radeksimko:
> Could you describe a bit more the LS workflow

The LS currently doesn't install Terraform, and I'm not sure _when_ it will or _how_ yet, but I'm assuming that we'd prompt the user to confirm the installation. In order to do that we'll need more flexibility around composing different parts of the library.

I imagine the workflow would be something like

1. Is there any Terraform installed? (filesystem discovery)
2. What are the version constraints for the code being edited? (LS will parse `terraform` blocks)
3. Is that discovered version compatible with these constraints?
4. If not, what is the latest available version within these constraints? (Checkpoint discovery)
5. Present a prompt to the user "No terraform found in PATH, do you want to install X.Y.Z for /path/to/dir?"
6. Does the user consent with the installation of version `X.Y.Z`?
   - they may cancel, and install Terraform via APT/Homebrew/... and/or make sure their OS-wide installation is discoverable in `$PATH`
7. Install version `X.Y.Z`. (installation)

> why it wouldn't be sufficient to do tfdisco.Find(tfdisco.LookPath()) earlier, and tfdisco.Find(tfdisco.LatestVersion()) (for instance) later?

It's not that it's impossible, it's more the UX/API design that makes it difficult to expand and support more use cases, because it assumes that everyone will need both discovery and installation at the same time, which I don't think is true.

Terraform can be installed outside of `terraform-exec` entirely and that method of installation may even be preferred in situations where there is practical access to OS package manager and user is in control (as opposed to CI or other automation-related scenarios).

The discovery runs `terraform version`, but doesn't provide the output to the consumer, making them run it again.

`tfdisco.LatestVersion()`(without more closely inspecting the argument names or implementation) makes me think it will just discover the latest version and return it either as `go-version.Version` or as a string. That is why I think it belongs to a different package, or at least should be renamed to `InstallLatestVersion` and `ExactVersion` to `InstallExactVersion`.

--- 

I don't want to overburden you with all these different scenarios at such early stage though, because I think the initial implementation can be simple and doesn't need to satisfy all different use cases, but I do think that we should leave some room in the API so we can tweak it more easily to cater more use cases.

With that in mind I'd be happy for the "find or install" logic to remain in the library, but it should be clear(er) upfront that this is what it does.

Maybe you could turn `Find` into `FirstExecutablePath` and `*findOption` into an interface
```go
type ExecutablePathFinder interface {
    ExecPath() string
}
```
and make each installer and discoverer comply with that interface?

Then you'd use it like this:

```go
FirstExecutablePath(
    tfdisco.ExactPath("/path/to/tf"),
    tfdisco.LookPath(),
    tfinstall.NewInstaller("latest"),
    tfinstall.NewInstaller("0.12.28"))
```
which leaves room for API surface growth on installer side, discovery side and the "coupling" side.

kmoe:
I've reverted the package name to `tfinstall`, because the majority of use cases are concerned with installing or otherwise obtaining a Terraform binary. The options pattern is now similar to that in `tfexec` itself, along the lines described by @radeksimko. This should make the usage clearer, and also improve the structure of the GoDoc, so `tfinstall.LatestVersion()` etc are clearly intended as function options (see the GoDoc for `tfexec`, https://godoc.org/github.com/hashicorp/terraform-exec/tfexec).

PR #12, opened 2020-07-03
CI and release process

kmoe:
fixes #11 

Based on the SDK CircleCI setup.

### PR builds

On each PR, a matrix of builds is run:
- build and test with Go 1.12
- build and test with Go 1.14
- gofmt and go vet checks (Go 1.14)
- build and test on Windows

There is a macOS executor on CircleCI, but it doesn't have Go installed, and the architecture is sufficiently similar to Linux to leave this out for now. If we want to test all PRs on macOS we can add this later.

### Release builds

As in the SDK, to release from `master` one needs only to update CHANGELOG.md to include a `#x.y.z (Unreleased)` line and then trigger the `trigger_release` build.

The release workflow will then:
 - run all build and test jobs as above
 - format the changelog
 - modify the version in `tfinstall/version.go` (this will be removed in https://github.com/hashicorp/terraform-exec/issues/22)
 - commit and push changes as a bot user
 - tag the version

Git commits and tags are signed using the bot's GPG key.

The release workflow was tested by temporarily adding `ci-release` to its allowed branches, preparing the branch for a v0.0.1 release, and verifying the changes committed and pushed to this branch by the bot.

PR #16, opened 2020-07-06
Remove discovery of terraform executable from tfexec

kmoe:
The `tfinstall` package introduced in https://github.com/hashicorp/terraform-exec/pull/8 is intended to replace `tfexec`'s automatic discovery of a `terraform` binary on the system PATH.

We also introduce the use of `tfinstall` in `tfexec`'s tests in order to run these tests, which require `terraform`, in CI (#12).

Please ignore the failing CI on this PR - it's not set up yet properly and won't be until #12 is merged.


tfexec/terraform.go
```diff
@@ -35,10 +35,8 @@ func NewTerraform(workingDir string, execPath string) (*Terraform, error) {
 	}
 
 	if execPath == "" {
-		execPath, err = FindTerraform()
-		if err != nil {
-			return nil, &ErrNoSuitableBinary{err: err}
-		}
+		err := fmt.Errorf("NewTerraform: please supply the path to a Terraform executable using execPath, e.g. using the tfinstall package.")
+		return nil, &ErrNoSuitableBinary{err: err}
```
paultyng, line 39: It may be useful to do this as `fmt.Errorf("...: %w", &ErrNoSuitableBinary{})` because it will force callers in to the `.Is` / `.As` consumption model.

tfexec/terraform_test.go
```diff
@@ -19,11 +20,30 @@ const testConfigFileName = "main.tf"
...
+
+func TestMain(m *testing.M) {
+	var err error
+	td, err := ioutil.TempDir("", "tfinstall")
+	if err != nil {
+		panic(err)
+	}
+	defer os.RemoveAll(td)
+
+	tfPath, err = tfinstall.Find(tfinstall.LookPath(), tfinstall.LatestVersion(td, true))
```
paultyng, line 33: You may want to add a finder here that takes an exact version from an env var if specified, so we can more easily matrix test this like `terraform-plugin-test`

kmoe: Thanks, this is a good suggestion for #19 - we won't need it until then.

tfexec/terraform.go
kmoe: The error doesn't need to be wrapped for `.As()` to work - I've added a unit test which tests this.

PR #24, opened 2020-07-11
Fix addr and ID usage for import

paultyng:
Also, fix minor issue of orphaned installed directories.

This is dependent upon #23


tfexec/import.go
```diff
@@ -75,8 +67,8 @@ func (opt *VarFileOption) configureImport(conf *importConfig) {
 	conf.varFile = opt.path
 }
 
-func (t *Terraform) Import(ctx context.Context, opts ...ImportOption) error {
-	importCmd := t.ImportCmd(ctx, opts...)
+func (t *Terraform) Import(ctx context.Context, address, id string, opts ...ImportOption) error {
```
paultyng, line 70: I have these as explicit strings here instead of the options model since they are required for this command.

paultyng:
Need to add a runtime test

kmoe:
Failing test:

```
--- FAIL: TestCheckpointDisablePropagation (0.11s)
    terraform_test.go:58: expected command env to be [FOOBAR=1 CHECKPOINT_DISABLE=1 TF_LOG=], but it was [CHECKPOINT_DISABLE=1 FOOBAR=1 TF_LOG=]
FAIL
```

I think this test is flaky but this is the first time I've caught it in a build. Will fix separately.

tfexec/import.go
kmoe: This is reasonable, but it means a breaking change for tfexec if `terraform import` gains or loses required arguments during future `terraform` versions.

To my knowledge there is no other `terraform` subcommand with required arguments, but I may be wrong...

tfexec/options.go
```diff
@@ -40,6 +32,11 @@ func Backup(path string) *BackupOption {
 	return &BackupOption{path}
 }
 
+// DisableBackup is a convenience method for Backup("-"), indicating backup state should be disabled.
+func DisableBackup() *BackupOption {
```
kmoe, line 36: Happy with this direction - we can provide more convenience options as use cases become clear.

PR #25, opened 2020-07-12
Stop exporting exec.Cmd versions of methods

paultyng:
We should only export the methods that execute commands to force all compatibility needs through the controlled interfaces (options, etc), as opposed to letting consumers further customize things like environment. This defeats the purpose of having a canonical testing interface for programmatic CLI invocation.

(per the RFC discussion)

This is based on #23


kmoe:
Thanks; please rebase #7, #23, #24, and #25 so that CI can be run.

PR #29, opened 2020-07-13
tfinstall cmd

### LINKED ISSUE #15: tfinstall CLI utility
The `tfinstall` package can also provide a `tfinstall` binary for quickly installing Terraform, e.g. as a CI build step. Probably best as an `internal/cmd` package so it's not confused with the library's public API.

Example CLI:

```
tfinstall [-latest-version] [-version=version] OUTDIR
```

kmoe:
closes #15 

First pass at a `tfinstall` binary. Current usage:

```
Usage: tfinstall [--dir=DIR] VERSION

  Downloads, verifies, and installs a terraform binary of version
  VERSION from releases.hashicorp.com. VERSION must be a valid version under
  semantic versioning, or "latest".

  If a binary is successfully installed, its path will be printed to stdout.

  Unless --dir is given, the default system temporary directory will be used.

Options:
  --dir          Directory into which to install the terraform binary. The
                 directory must exist.

Examples:
  tfinstall 0.12.28
  tfinstall latest
  tfinstall 0.13.0-beta3
  tfinstall --dir=/home/kmoe/bin 0.12.28
```

@paultyng @radeksimko looking for feedback on the UX here. This is intended as a simple utility for use in CI build steps. When we're happy with it I can add config to build and release binaries as part of the release process.

Implemented using a somewhat bastardised version of mitchellh/cli, because I wanted `cli.Ui` but there's currently no way to have flags without subcommands (https://github.com/mitchellh/cli/issues/24).


paultyng:
The UX looks good to me, the only thing I would probably prefer would that no args/flags at all (ie. just `tfinstall`) is equivalent to `tfinstall latest`.

cmd/tfinstall/main.go
```diff
@@ -0,0 +1,106 @@
...
+	filter := &logutils.LevelFilter{
+		Levels:   []logutils.LogLevel{"DEBUG", "WARN", "ERROR"},
+		MinLevel: logutils.LogLevel("WARN"),
+		Writer:   os.Stderr,
+	}
+	log.SetOutput(filter)
+
+	ui := &cli.ColoredUi{
+		OutputColor: cli.UiColorBlue,
+		InfoColor:   cli.UiColorGreen,
```
radeksimko, line 24: Nitpick: Red for errors and yellow for warnings is common, but is there any particular reason not to keep `Output` and `Info` as default (probably white)? I think that too many colours then defeat the purpose of colouring (catching attention) and errors and warnings get lost.

kmoe: Thanks, that does make more sense here - I copied these from tf-sdk-migrator, which has more complex output.

kmoe:
@paultyng this is ready for further work now. I'll add the goreleaser config (https://github.com/hashicorp/terraform-exec/issues/71).

paultyng:
Just rebased and `go mod tidy` on this in my update.

PR #32, opened 2020-07-18
Env Var handling and TF_LOG_PATH support

paultyng:
This is a WIP to modify env var handling slightly and add support for TF_LOG_PATH to tfexec. Unsure if this fulfills the existing use case yet, but just putting this up for further discussion in the meantime.

### DIFF
--- a/tfexec/terraform_test.go
+++ b/tfexec/terraform_test.go
@@ -86,7 +86,7 @@ func TestCheckpointDisablePropagation(t *testing.T) {
 		"FOOBAR": "1",
 	})
 	initCmd := tf.initCmd(context.Background())
-	expected := []string{"CHECKPOINT_DISABLE=1", "FOOBAR=1", "TF_LOG="}
+	expected := []string{"CHECKPOINT_DISABLE=1", "FOOBAR=1", "TF_IN_AUTOMATION=1", "TF_LOG=", "TF_LOG_PATH="}
 	s := initCmd.Env
 	sort.Strings(s)
 	actual := s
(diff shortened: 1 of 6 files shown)


tfexec/terraform_test.go
```diff
@@ -56,12 +96,15 @@ func TestCheckpointDisablePropagation(t *testing.T) {
...
+	err = tf.SetEnv(map[string]string{
 		"CHECKPOINT_DISABLE": "",
 		"FOOBAR":             "1",
 	})
+	if err != nil {
+		t.Fatal(err)
+	}
 	initCmd = tf.initCmd(context.Background())
-	expected = []string{"CHECKPOINT_DISABLE=", "FOOBAR=1", "TF_LOG="}
+	expected = []string{"CHECKPOINT_DISABLE=", "FOOBAR=1", "TF_INPUT=0", "TF_IN_AUTOMATION=1", "TF_LOG=", "TF_LOG_PATH="}
```
kmoe, line 107: Why replace `-input=false` with the `TF_INPUT` env var? I think the former is more explicit.

paultyng: its safer from a compatibility standpoint, if `-input` were added later to a command (or if we didn't historically audit its availability to confirm its available in all versions in all commands necessary), we wouldn't need to version check to see whether we need to pass a flag or an env var, can just always send an env var with no worries about usage errors.

kmoe: This means that instead of an explicit failure in the case where `-input=false` is not supported but we still attempt to use it, we have a silent failure where `TF_INPUT` is not supported but we still attempt to use it.

paultyng: The failure mode for not supplying it is a hanging execution, on the user machine, as its specific to the config being run on, not the program using this.

It was already missed in the destroy methods, this caused a runtime hang for me in something I was testing, not a development time problem, I'm erring on the side of not hanging user machines. vs erring on the side of an extra flag syntax issue.

paultyng:
@kmoe tests look green now.

PR #43, opened 2020-08-06
End to end testing

### LINKED ISSUE #19: End-to-end test framework
Test the non-`*Cmd` versions of the `tfexec` functions, preferably with as many test cases as possible.

Borrow test cases from Core and terraform-json.

Ideally perform some sort of matrix testing across multiple Terraform versions and OS environments (needs #11).

kmoe:
Adds new `e2etest` package and an initial set of test cases. This is by no means complete, but seems a sensible point at which to merge the long-running `end-to-end-testing` branch. Further test cases should be added as their respective implementations are touched.

Closes #19.


tfexec/errors.go
```diff
@@ -19,6 +24,8 @@ func parseError(stderr string) error {
 	default:
 		return errors.New(stderr)
 	}
+
+	return fmt.Errorf("error from Terraform CLI:\n%s", stderr)
```
paultyng, line 28: This is unreachable since there is a `default` case.
```suggestion
```

kmoe: Thanks, I think that was a bad rebase!

PR #46, opened 2020-08-07
Add management of TF_APPEND_USER_AGENT

### LINKED ISSUE #9: Support passing User-Agent
Tools executing Terraform (such as Terraform Language Server) would benefit from measuring impact/usage and one way of doing it would be by appending an identifier to User-Agent by passing `TF_APPEND_USER_AGENT` which then enables filtering data from whatever upstream backend requests were sent to.

https://github.com/hashicorp/terraform-plugin-test/issues/22

I'm not sure if this should be a first-class config option, but I'm creating this issue to discuss _whether_ it should.

paultyng:
Fixes #9

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -2,10 +2,13 @@ package tfexec
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"strings"
+
+	"github.com/hashicorp/terraform-exec/internal/version"
 )
 
 const (
@@ -15,6 +18,7 @@ const (
 	automationEnvVar        = "TF_IN_AUTOMATION"
 	logPathEnvVar           = "TF_LOG_PATH"
 	reattachEnvVar          = "TF_REATTACH_PROVIDERS"
+	appendUserAgentEnvVar   = "TF_APPEND_USER_AGENT"
 
 	varEnvVarPrefix = "TF_VAR_"
 )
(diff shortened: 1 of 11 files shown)


tfexec/cmd.go
```diff
@@ -74,6 +79,14 @@ func (tf *Terraform) buildEnv(mergeEnv map[string]string) []string {
 		env[checkpointDisableEnvVar] = os.Getenv(checkpointDisableEnvVar)
 	}
 
+	// always override user agent
+	ua := mergeUserAgent(
+		os.Getenv(appendUserAgentEnvVar),
+		tf.appendUserAgent,
+		fmt.Sprintf("HashiCorp-terraform-exec/%s", version.ModuleVersion()),
```
radeksimko, line 86: Shouldn't these values be in reverse order? i.e. static `HashiCorp-terraform-exec` first and then other values being _appended_, rather than prepended?

paultyng: I think it depends. arguably "terraform cloud" is more significant / distinct than terraform exec I think in the case of that system? As there will be many terraform exec apps, so the app itself is the more significant identifier?

radeksimko: I don't disagree, but the problem is that the ENV variable is called _append_, not _prepend_, so we would need to reflect it there somehow (with another/changed variable), else the behaviour won't match the name.

paultyng: Thats the common convention in programming but not the english definition of the word. It was a poor original choice for the env var, but its what we have so no option to change it now.

radeksimko (approved):

PR #48, opened 2020-08-14
Add better version matrix handling and additional version testing for all e2e tests

paultyng:
In addition to TF version matrix updates, added go 1.15, changed 1.12 and 1.13 to building only (not testing). Testing only in 1.14 and 1.15 as they are the support Go versions currently.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,26 +30,27 @@ jobs:
       - win_install_go
       - go_test
 
-  "go112_build":
-    docker:
-      - image: circleci/golang:1.12
-    steps:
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -86,50 +104,74 @@ workflows:
...
-      - wintest
-      - "go112_build"
-      - "go112_test"
-      - "go114_build"
-      - "go114_test":
+      - wintest:
+          requires:
+            - winbuild
+
+      # build only for these versions
```
kmoe, line 111: :+1:

Issue #45, opened 2020-08-06
Add ability to install latest from Terraform git repo


paultyng:

paultyng:
We should add a cron testing against `refs/heads/master` as well, at least daily, for early warning of breakage.

PR #51, opened 2020-08-16
Add support for finding / installing Terraform by git ref

paultyng:
This is a breaking change due to adding a `context.Context` to `tfinstall.Find`

This really balloons up testing time, we should think through any other strategies to possibly speed that up?


paultyng:
Unclear why these won't execute in Circle, builds seem to pass locally.

tfinstall/download.go
```diff
@@ -0,0 +1,133 @@
...
+	tfDir, err := ensureInstallDir(installDir)
+	if err != nil {
+		return "", err
+	}
+
+	// setup: getter client
+
+	// TODO: actually use this header...
+	// httpHeader := make(http.Header)
+	// httpHeader.Set("User-Agent", "HashiCorp-tfinstall/"+Version)
```
paultyng, line 42: @kmoe I noticed the header doesn't actually seem to be attached anywhere, but haven't gotten a chance to look at how to set it in go-getter yet. Just an FYI if you happen to poke around in here.

go.mod
```diff
@@ -4,10 +4,11 @@ go 1.14
 
 require (
 	github.com/davecgh/go-spew v1.1.1
+	github.com/go-git/go-git/v5 v5.1.0
```
kmoe, line 7: Why not use go-getter to get the git repo?

kmoe:
These tests are a bit much for local development, but we do want to run them in CI (both PR and release workflows). An env var to switch all the `master` tests on/off would be sensible.

go.mod
paultyng: I thought about it. go-getter uses CLI execution of `git`, and a lot of code around SSH key handling and other things which I felt was overkill. I guess this strategy is already requiring Go on your PATH, but trying to keep the requirements minimal.

paultyng:
Yeah, I will remove these from the defaults, but let you opt in to running master as "latest 0.13" or something with an env var.

tfexec/internal/e2etest/util_test.go
```diff
@@ -18,29 +19,34 @@ import (
...
 	t.Helper()
-	runTestVersions(t, []string{
+
+	versions := []string{
 		testutil.Latest011,
 		testutil.Latest012,
 		testutil.Latest013,
-	}, fixtureName, cb)
+	}
+	if override := os.Getenv("TFEXEC_E2ETEST_VERSIONS"); override != "" {
```
paultyng, line 32: @kmoe I added this override, and I think we should maybe just run a cron CI run against master instead of on PR, but I'm not sure how to set it up in Circle, so will follow up in a future PR with that.

paultyng:
This partially solves #45, but we should also turn on cron based runs against it.

tfexec/internal/e2etest/util_test.go
kmoe: added https://github.com/hashicorp/terraform-exec/issues/70 for this

kmoe (approved):

PR #54, opened 2020-08-16
add ShowState, ShowPlan

kmoe:
`terraform show -json PATH` will output the state file or plan file at `PATH` in the corresponding `tfjson` format. 

Since we must know ahead of time whether we're returning `*tfjson.State` or `*tfjson.Plan`, we now require the user to call either `ShowState()` or `ShowPlan()` if they are supplying a `PATH`.

Add end-to-end tests. Negative compatibility cases will be covered in https://github.com/hashicorp/terraform-exec/issues/58.


tfexec/show.go
```diff
@@ -10,6 +10,8 @@ import (
 	tfjson "github.com/hashicorp/terraform-json"
 )
 
+// Show reads the default state path and outputs the state.
+// To read a state or plan file, ShowState or ShowPlan must be used instead.
 func (tf *Terraform) Show(ctx context.Context) (*tfjson.State, error) {
```
paultyng, line 15: Should we just remove this version?

kmoe: No, `terraform show -json` is a valid use case. If `Show()` were removed, consumers would have to call either:
 - `ShowState(ctx, "terraform.tfstate")`, exposing the name of the state file, which is an implementation detail, or 
 - `ShowState(ctx, "")`, which users would expect to call `terraform show -json ""`, but would in fact be a special case calling `terraform show -json`, since the former leads to a CLI error.

kmoe:
@paultyng any idea why the test fails on Windows?

tfexec/show.go
paultyng: Makes sense, I wonder if the other two methods should be called `ShowPlanFile`, `ShowStateFile`, i just worry it could be confusing as to what they are calling under the hood and why they are differentiated.

paultyng:
I wonder if we should just write up a `runTerraformCmdJSON` as we have a lot of these stdout unmarshalling patterns.

tfexec/show.go
paultyng: Should you error out if they pass a blank?

kmoe:
`runTerraformCmdJSON()` could work, but we're marshalling into different JSON types each time...

paultyng:
yeah, you would pass the pointer just like `json.Unmarshal`

PR #55, opened 2020-08-18
Add -detailed-exitcode option to Plan

sudomateo:
This adds the `-detailed-exitcode` option to `tfexec.Plan()` and returns
a boolean detailing whether a plan diff is empty or non-empty.

This allows the caller to make decisions when the plan has changes.

```
hasChanges, err := tf.Plan(context.Background())
if err != nil {
  panic(err)
}
if hasChanges {
  // Do something when the plan has changes.
}
```

### DIFF
--- a/tfexec/plan.go
+++ b/tfexec/plan.go
@@ -8,25 +8,27 @@ import (
 )
 
 type planConfig struct {
-	destroy     bool
-	dir         string
-	lock        bool
-	lockTimeout string
-	out         string
-	parallelism int
-	refresh     bool
-	state       string
-	targets     []string
-	vars        []string
-	varFile     string
+	destroy          bool
+	detailedExitCode bool
+	dir              string
+	lock             bool
+	lockTimeout      string
+	out              string
+	parallelism      int
+	refresh          bool
+	state            string
+	targets          []string
+	vars             []string
+	varFile          string
 }
 
 var defaultPlanOptions = planConfig{
-	destroy:     false,
-	lock:        true,
-	lockTimeout: "0s",
-	parallelism: 10,
-	refresh:     true,
+	destroy:          false,
+	detailedExitCode: false,
+	lock:             true,
+	lockTimeout:      "0s",
+	parallelism:      10,
+	refresh:          true,
 }
 
 type PlanOption interface {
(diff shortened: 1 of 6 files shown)


paultyng:
Per the internal conversation, I think we should make the following changes to surface this functionality:

* always pass `-detailed-exitcode` for `Plan`
* change the `Plan` function signature to `Plan(...) (bool, error)` where the bool represents if a diff is present or not (exit code 0 or 2)

We probably need a test case for a diff and no diff, not sure if that is present in the e2e testing yet, so probably need to add a test case that modifies the TF file, I can assist with this though if you have any questions.

sudomateo:
@paultyng @kmoe this is ready for review. Thank you for the feedback!

tfexec/plan.go
```diff
@@ -77,8 +77,21 @@ func (opt *DestroyFlagOption) configurePlan(conf *planConfig) {
...
+//
+// The returned boolean is false when the plan diff is empty (no changes) and
+// true when the plan diff is non-empty (changes present).
+//
+// The returned error is nil if `terraform plan` has been executed and exits
+// with either 0 or 2.
+func (tf *Terraform) Plan(ctx context.Context, opts ...PlanOption) (bool, error) {
+	cmd := tf.planCmd(ctx, opts...)
+	err := tf.runTerraformCmd(cmd)
+	if err != nil && cmd.ProcessState.ExitCode() == 2 {
```
paultyng, line 91: You probably need to confirm this is `ExitError`, ie something like: 
```suggestion
	var exitErr *exec.ExitError
	if errors.As(err, &exitErr) && cmd.ProcessState.ExitCode() == 2 {
```

sudomateo: I don't believe this will work since the `err` here is returned from `parseError()` (hitting the `default` case). At that time the original `err` is essentially discarded.

sudomateo: Should I perhaps add another case to `parseError()` that checks the original error and returns a new error type that I can then check from `Plan()`?

paultyng: yeah, i think we should probably fix the default case to just wrap the `ExitError` so that its compatible with `errors.As`/`Is`.

paultyng:
This needs a rebase but I think we just merge it as is. The refactor for the error handling won't be user facing, so can just be a future improvement if someone gets to it.

PR #56, opened 2020-08-19
Add terraform workspace commands

paultyng:


tfexec/internal/e2etest/testdata/workspaces/terraform.tfstate.d/foo/.keep
```diff
@@ -0,0 +1 @@
+this file is used to hold this directory in git
```
paultyng, line 1: directory presence is used by the Terraform CLI for workspaces

paultyng:
I can move the first 3 commits to a separate PR since they are probably easier to merge, let me do that.

tfexec/errors.go
```diff
@@ -44,9 +46,13 @@ func parseError(err error, stderr string) error {
 		return &ErrNoInit{stderr: stderr}
 	case noConfigErrRegexp.MatchString(stderr):
 		return &ErrNoConfig{stderr: stderr}
-	default:
-		return errors.New(stderr)
+	case workspaceDoesNotExistRegexp.MatchString(stderr):
+		submatches := workspaceDoesNotExistRegexp.FindStringSubmatch(stderr)
+		if len(submatches) == 2 {
```
appilon, line 51: This feels odd to me but I have no context.

paultyng: Yeah I think at some point we will need to overhaul this from a `switch` to something a bit more sophisticated.

paultyng: I could use named groups and make it a bit cleaner possibly, but not without a lot of extra code for this single case.

tfexec/options.go
```diff
@@ -55,6 +55,18 @@ func Config(path string) *ConfigOption {
 	return &ConfigOption{path}
 }
 
+// CopyStateOption represents the -state flage for terraform workspace new. This flag is used
```
appilon, line 58:
```suggestion
// CopyStateOption represents the -state flag for terraform workspace new. This flag is used
```

tfexec/options.go
```diff
@@ -55,6 +55,18 @@ func Config(path string) *ConfigOption {
...
+// to copy an existing state file in to the new workspace.
+type CopyStateOption struct {
+	path string
+}
+
+// CopyState represents the -state flage for terraform workspace new. This flag is used
```
appilon, line 64:
```suggestion
// CopyState represents the -state flag for terraform workspace new. This flag is used
```

PR #57, opened 2020-08-20
Add error for missing variable

paultyng:
This only tests plans, I assumed the error strings were the same, but probably wouldn't hurt to add an apply test as well.

I also resolved a minor issue where erroring commands were not in the test logging.

### DIFF
--- a/tfexec/errors.go
+++ b/tfexec/errors.go
@@ -5,6 +5,20 @@ import (
 	"fmt"
 	"os/exec"
 	"regexp"
+	"strings"
+)
+
+var (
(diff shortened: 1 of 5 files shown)


tfexec/errors.go
```diff
@@ -5,6 +5,20 @@ import (
 	"fmt"
 	"os/exec"
 	"regexp"
+	"strings"
+)
+
+var (
+	// The "Required variable not set:" case is for 0.11
```
paultyng, line 12: Moved these out here so that the bad regexp panic happens at `init` time, and not at error parse time.

kmoe (approved):

PR #64, opened 2020-08-31
terraform console

paultyng:
This does not currently work in Windows (see https://github.com/hashicorp/terraform/issues/18242). For now, just an error if invoked from Windows, if support is added, we can change that to a compatibility check.

### DIFF
--- a/tfexec/console.go
+++ b/tfexec/console.go
@@ -1,0 +1,56 @@
+package tfexec
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"runtime"
+	"strings"
+)
+
+// Console represents the console subcommand to the Terraform CLI.
+func (tf *Terraform) Console(ctx context.Context, expression string) (string, error) {
+	if runtime.GOOS == "windows" {
+		return "", fmt.Errorf("terraform console does not support stdin scripting on Windows currently, see https://github.com/hashicorp/terraform/issues/18242")
+	}
+
+	// TODO: [DIR] param option
+	// TODO: var option
+	// TODO: var-file option
+	cmd := tf.buildTerraformCmd(ctx, "console")
+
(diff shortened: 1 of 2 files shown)


paultyng:
The windows handling is fixed upstream so this should work on windows in 0.15+

radeksimko (changes requested):
Was there any particular use case you had in mind for this command? I think we might eventually need it in Terraform LS, although I can't say for certain at this point.

Regardless of the needs, the output from `terraform console` actually may _not_ always be valid HCL syntax.

With that in mind I can imagine two ways forward in the near future:

1. We can close this PR unmerged and reopen the conversation later, when we have well-defined use cases for it - at which point we could also reopen the conversation about `console` supporting JSON output
2. Reimplement `ConsoleJSON` as `Console` which returns plain stdout as-is, review and merge.

I'm leaning more towards (1) unless there is a clear use case for this, but I'm interested to hear your opinion @paultyng

paultyng:
I thought this may be an easier way to do state inspection or things like testing / consul-tf-sync, without needing to worry about outputs, but I stopped pursuing it once I hit the windows stumbling block. It may be worth picking up once 1.0 is out, but until then its still just going to be broken for a decent portion of the audience, so you could probably just close.

radeksimko:
> but until then its still just going to be broken for a decent portion of the audience, so you could probably just close.

We could just gate it on TF 0.15+ and `GOOS` = `windows` or something.

I will ask the TF SDK and Consul teams about the needs and use cases and decide based on that.

Thanks for sharing the context!

radeksimko:
Both TF SDK and relevant Consul team told me that they don't have need for `terraform console` integration at this point, so I'm going to close this PR, but as mentioned I'd be happy to revisit once we have more specific use case in mind, so that we can also design the API around it.

Thank you for the effort invested into the PR.

PR #68, opened 2020-09-01
terraform validate support

paultyng:
The types don't seem to be in terraform-json yet, so need to add those there first.


tfexec/validate_types.go
```diff
@@ -0,0 +1,30 @@
+package tfexec
+
+// TODO: move these types to terraform-json
```
appilon, line 3: Is this project 1.0? Would it be a breaking change to export these types, and then move them to tfjson and remove them here?

radeksimko: I have just raised https://github.com/hashicorp/terraform-json/pull/23 - will see if I can get it merged 🔜

kmoe (approved):
Good stuff. This can go in once https://github.com/hashicorp/terraform-json/pull/23 is merged, so we're not duplicating the types. @radeksimko, https://github.com/hashicorp/terraform-json/pull/23 LGTM but I think the owners of that repo should approve.

PR #78, opened 2020-09-07
add ReattachOption

kmoe:
Binary testing reattach functionality requires the `TF_PROVIDERS_REATTACH` env var to be present in the environment in which the Terraform CLI is run. In normal usage, such as in terraform-plugin-test, the value of this env var will change during the lifetime of a `*Terraform`. We therefore provide a way of setting the reattach info string on every relevant Terraform CLI command as if it is an option.

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -114,12 +114,20 @@ func (tf *Terraform) buildEnv(mergeEnv map[string]string) []string {
 	// force usage of workspace methods for switching
 	env[workspaceEnvVar] = ""
 
+	if tf.disablePluginTLS {
+		env[disablePluginTLSEnvVar] = "1"
+	}
+
+	if tf.skipProviderVerify {
+		env[skipProviderVerifyEnvVar] = "1"
+	}
+
 	return envSlice(env)
 }
 
-func (tf *Terraform) buildTerraformCmd(ctx context.Context, args ...string) *exec.Cmd {
+func (tf *Terraform) buildTerraformCmd(ctx context.Context, mergeEnv map[string]string, args ...string) *exec.Cmd {
 	cmd := exec.CommandContext(ctx, tf.execPath, args...)
-	cmd.Env = tf.buildEnv(nil)
+	cmd.Env = tf.buildEnv(mergeEnv)
 	cmd.Dir = tf.workingDir
 
 	tf.logger.Printf("[INFO] running Terraform command: %s", cmdString(cmd))
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -15,14 +15,16 @@ import (
 )
 
 const (
-	checkpointDisableEnvVar = "CHECKPOINT_DISABLE"
-	logEnvVar               = "TF_LOG"
-	inputEnvVar             = "TF_INPUT"
-	automationEnvVar        = "TF_IN_AUTOMATION"
-	logPathEnvVar           = "TF_LOG_PATH"
-	reattachEnvVar          = "TF_REATTACH_PROVIDERS"
-	appendUserAgentEnvVar   = "TF_APPEND_USER_AGENT"
-	workspaceEnvVar         = "TF_WORKSPACE"
+	checkpointDisableEnvVar  = "CHECKPOINT_DISABLE"
+	logEnvVar                = "TF_LOG"
+	inputEnvVar              = "TF_INPUT"
+	automationEnvVar         = "TF_IN_AUTOMATION"
+	logPathEnvVar            = "TF_LOG_PATH"
+	reattachEnvVar           = "TF_REATTACH_PROVIDERS"
+	appendUserAgentEnvVar    = "TF_APPEND_USER_AGENT"
+	workspaceEnvVar          = "TF_WORKSPACE"
+	disablePluginTLSEnvVar   = "TF_DISABLE_PLUGIN_TLS"
+	skipProviderVerifyEnvVar = "TF_SKIP_PROVIDER_VERIFY"
 
 	varEnvVarPrefix = "TF_VAR_"
 )
@@ -35,6 +37,8 @@ var prohibitedEnvVars = []string{
 	reattachEnvVar,
 	appendUserAgentEnvVar,
 	workspaceEnvVar,
+	disablePluginTLSEnvVar,
+	skipProviderVerifyEnvVar,
 }
 
 func envMap(environ []string) map[string]string {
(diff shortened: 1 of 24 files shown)


tfexec/options.go
```diff
@@ -172,6 +172,14 @@ func PluginDir(pluginDir string) *PluginDirOption {
 	return &PluginDirOption{pluginDir}
 }
 
+type ReattachOption struct {
+	info string
+}
+
+func Reattach(reattach string) *ReattachOption {
```
paultyng, line 179: this should probably take the structure, not a string, similar to all our other terraform-json handling

tfexec/terraform.go
```diff
@@ -129,6 +133,18 @@ func (tf *Terraform) SetAppendUserAgent(ua string) {
 	tf.appendUserAgent = ua
 }
 
+// SetDisablePluginTls sets the TF_DISABLE_PLUGIN_TLS environment variable for
+// Terraform CLI execution.
+func (tf *Terraform) SetDisablePluginTls(disabled bool) {
```
paultyng, line 138: You can probably just make this return an error for compatibility mismatch that people can just ignore if they want.

tfexec/terraform.go
```diff
@@ -129,6 +133,18 @@ func (tf *Terraform) SetAppendUserAgent(ua string) {
...
+	tf.disablePluginTls = disabled
+}
+
+// SetSkipProviderVerify sets the TF_SKIP_PROVIDER_VERIFY environment variable
+// for Terraform CLI execution.
+func (tf *Terraform) SetSkipProviderVerify(skip bool) {
```
paultyng, line 144: You can probably just make this return an error for compatibility mismatch that people can just ignore if they want.

paultyng:
Some comments inline, also need a compatibility check for 0.12.26 or greater

tfexec/terraform.go
kmoe: `TF_SKIP_PROVIDER_VERIFY` was added in v0.10.0-beta2, so no compatibility check needed: https://github.com/hashicorp/terraform/commit/9944ea6886e8d4da1190b93f6d68088b9468c874

paddycarver:
I agree with all @paultyng's comments, and have no other comments of my own to add. :+1:

tfexec/terraform.go
paultyng: But it's no longer used in 0.13

tfexec/cmd.go
```diff
@@ -63,10 +67,12 @@ func envSlice(environ map[string]string) []string {
 }
 
 func (tf *Terraform) buildEnv(mergeEnv map[string]string) []string {
+	osEnv := envMap(os.Environ())
```
paultyng, line 70: you should restore this to the old code, since its not always necessary, and this invokes a syscall even if thrown away and some mutex's, etc.

tfexec/cmd.go
```diff
@@ -15,14 +15,16 @@ import (
...
-	workspaceEnvVar         = "TF_WORKSPACE"
+	checkpointDisableEnvVar  = "CHECKPOINT_DISABLE"
+	logEnvVar                = "TF_LOG"
+	inputEnvVar              = "TF_INPUT"
+	automationEnvVar         = "TF_IN_AUTOMATION"
+	logPathEnvVar            = "TF_LOG_PATH"
+	reattachEnvVar           = "TF_REATTACH_PROVIDERS"
+	appendUserAgentEnvVar    = "TF_APPEND_USER_AGENT"
+	workspaceEnvVar          = "TF_WORKSPACE"
+	disablePluginTlsEnvVar   = "TF_DISABLE_PLUGIN_TLS"
```
paultyng, line 26: initialisms should be all caps: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
```suggestion
	disablePluginTLSEnvVar   = "TF_DISABLE_PLUGIN_TLS"
```

tfexec/options.go
```diff
@@ -172,6 +176,39 @@ func PluginDir(pluginDir string) *PluginDirOption {
...
+type ReattachConfigAddr struct {
+	Network string
+	String  string
+}
+
+type ReattachOption struct {
+	info ReattachInfo
+}
+
+func marshalReattachString(info ReattachInfo) (string, error) {
```
paultyng, line 200: you should just hang this off the type to not add to the global namespace
```suggestion
func (info ReattachInfo) marshalReattachString() (string, error) {
```

tfexec/show.go
```diff
@@ -8,15 +8,44 @@ import (
...
+	reattachInfo ReattachInfo
+}
+
+var defaultShowOptions = showConfig{}
+
+type ShowOption interface {
+	configureShow(*showConfig)
+}
+
+func (opt *ReattachOption) configureShow(conf *showConfig) {
```
paultyng, line 21: Does `show` uses provider binaries? Not sure, I assumed it only used backend.

kmoe: Apparently it does something like that, because I got an error when I didn't supply the reattach env var...

tfexec/cmd.go
kmoe: Thanks, this was left over from an unused change

tfexec/terraform.go
```diff
@@ -125,8 +130,27 @@ func (tf *Terraform) SetLogPath(path string) error {
...
 // Terraform CLI execution.
-func (tf *Terraform) SetAppendUserAgent(ua string) {
+func (tf *Terraform) SetAppendUserAgent(ua string) error {
 	tf.appendUserAgent = ua
+	return nil
+}
+
+// SetDisablePluginTls sets the TF_DISABLE_PLUGIN_TLS environment variable for
+// Terraform CLI execution.
+func (tf *Terraform) SetDisablePluginTls(disabled bool) error {
```
paultyng, line 140: initialism: https://github.com/golang/go/wiki/CodeReviewComments#initialisms
```suggestion
func (tf *Terraform) SetDisablePluginTLS(disabled bool) error {
```

tfexec/terraform.go
```diff
@@ -125,8 +130,27 @@ func (tf *Terraform) SetLogPath(path string) error {
...
+	tf.disablePluginTls = disabled
+	return nil
+}
+
+// SetSkipProviderVerify sets the TF_SKIP_PROVIDER_VERIFY environment variable
+// for Terraform CLI execution.
```
paultyng, line 146:
```suggestion
// for Terraform CLI execution. This is no longer used in 0.13.0 and greater.
```

tfexec/show.go
paultyng: 👍

paultyng (approved):
minor comments inline, otherwise LGTM

Issue #6, opened 2020-06-12
Investigate / test tfenv usage


paultyng:
This is mostly relevant to the LS, but something we should investigate.

kmoe:
also possibly https://warrensbox.github.io/terraform-switcher/

kmoe:
See https://github.com/hashicorp/terraform-ls/issues/128 for unexpected behaviour cased by using a wrapper script to set appropriate env vars.

paultyng:
At least warn or let consumers know if these tools are being used, even if we aren't going to handle it.

kmoe:
Another one: https://github.com/Yleisradio/homebrew-terraforms

kmoe:
Results of investigation. Branch `tfenv-etc` contains some Dockerfiles for experimenting with these utilities in a clean environment (except homebrew-terraforms, which is only available for macOS).

## tfenv

Installing tfenv adds a shell script named `terraform` to the `$PATH`, so `which terraform` resolves to this script. This script checks whether `$TFENV_ROOT/version` is set, and errors if not. 

Currently, this means that any use of `tfinstall.LookPath()` will error and not fall through to the next ExecPathFinder, because it assumes that if it finds a `terraform` binary then it must report `terraform version` properly. Recommend relaxing this and will open PR to do so.

**`terraform` on path is now a shell script**: yes
**Terraform version may depend on working directory**: yes

## terraform-switcher

Installing a Terraform version via tfswitch replaces any existing Terraform binary with a symlink. This is only done when installing a new version of Terraform, so unlike tfenv there is no risk of the binary or script at `which terraform` not pointing to Terraform itself.

**`terraform` on path is now a shell script**: no
**Terraform version may depend on working directory**: yes

## homebrew-terraforms

The most conservative of the three helpers, this simply provides a `chtf` utility which allows the user to install and (with caching) switch between various versions of Terraform via the `brew` package manager. It therefore has no more impact on tfexec than brew itself.

**`terraform` on path is now a shell script**: no
**Terraform version may depend on working directory**: no

## Conclusion & recommendations

 - `tfinstall.LookPath()` - do not error out when an executable named `terraform` is found but `terraform version` does not return reasonable output; fall through to the next method instead.
 - Add note to README warning users of helpers that if the `terraform` on their path resolves to a different version depending on working directory, this will also be true of the Terraform executable located by `tfinstall.LookPath()`.
[...]

PR #79, opened 2020-09-10
Add environment variable helpers, prevent CLI env var usage

paultyng:
@kmoe i was adding prevention of usage of the CLI env vars and explored some convenience methods that may make usage of the `os.Environ` easier when desired.

Optionally, maybe we should just have a different method that uses `os.Environ` "cleanly", I'm not sure what the best UX is here, just trying it out as I had to make changes to how it looked up two types of prefixed env vars.

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -16,6 +16,7 @@ import (
 
 const (
 	checkpointDisableEnvVar  = "CHECKPOINT_DISABLE"
+	cliArgsEnvVar            = "TF_CLI_ARGS"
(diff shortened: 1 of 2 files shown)


kmoe:
This seems fine. I think there are some further env issues to solve before v1.0.0, in particular https://github.com/hashicorp/terraform-exec/issues/6 and its LS implications, so it's possible the env approach will change entirely, but I'd be happy to merge this now. I don't have a solution for the tfenv issues yet.

kmoe (approved):

PR #83, opened 2020-09-12
Add ShowPlanFileRaw

paddycarver:
The SDK used a non-JSON version of ShowPlanFile to return the
human-friendly, opaque-to-machines plan output during testing. With the
switch to tfexec, this is no longer possible, as all `terraform show`
commands have the `-json` flag applied, meaning all output will be JSON
output.

This adds a new function, `ShowRawPlanFile`, which returns a string of
the output of `terraform show` _without_ the `-json` flag, showing the
raw contents of the plan file. A new function was chosen to avoid
overloading the existing ShowPlanFile, and because the existing
ShowPlanFile returns a `tfjson.Plan` type, which we can't shoehorn a
non-JSON return type into easily. Rather than complicate the return
type, it felt like a new function was a better solution.

### DIFF
--- a/tfexec/show.go
+++ b/tfexec/show.go
@@ -1,6 +1,7 @@
 package tfexec
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os/exec"
@@ -45,7 +46,7 @@ func (tf *Terraform) Show(ctx context.Context, opts ...ShowOption) (*tfjson.Stat
 		mergeEnv[reattachEnvVar] = reattachStr
 	}
 
-	showCmd := tf.showCmd(ctx, mergeEnv)
+	showCmd := tf.showCmd(ctx, true, mergeEnv)
 
 	var ret tfjson.State
 	err = tf.runTerraformCmdJSON(showCmd, &ret)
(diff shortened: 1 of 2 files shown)


tfexec/show.go
```diff
@@ -146,8 +147,47 @@ func (tf *Terraform) ShowPlanFile(ctx context.Context, planPath string, opts ...
...
+		if err != nil {
+			return "", err
+		}
+		mergeEnv[reattachEnvVar] = reattachStr
+	}
+
+	showCmd := tf.showCmd(ctx, false, mergeEnv, planPath)
+
+	var ret bytes.Buffer
+	showCmd.Stdout = &ret
```
paultyng, line 175: This needs to use mergewriters just in case

paddycarver:
Updated to `ShowPlanFileRaw`, at @paultyng's suggestion, because he has opinions and I do not. :D

kmoe (approved):
LGTM. Note name of function `ShowPlanFileRaw` (looks like it was changed at some point) if you rebase and merge, as it uses the PR title.

tfexec/show.go
paddycarver: `mergeWriters` is already called by `tf.runTerraformCmd`, though? In fact, that's the _only_ command that calls it.

PR #88, opened 2020-09-24
fmt: Gate formatting command on 0.7.7

radeksimko:
This is mostly a UX enhancement which is already built into LS, so this is to help ease the transition.

### DIFF
--- a/tfexec/errors.go
+++ b/tfexec/errors.go
@@ -130,13 +130,8 @@ func (e *ErrVersionMismatch) Error() string {
 }
 
 func (e *ErrVersionMismatch) Is(target error) bool {
-	err, ok := target.(*ErrVersionMismatch)
-	if !ok {
-		return false
-	}
-	return err.Actual == e.Actual &&
-		err.MaxExclusive == e.MaxExclusive &&
-		err.MinInclusive == e.MinInclusive
+	_, ok := target.(*ErrVersionMismatch)
+	return ok
 }
 
 type ErrNoInit struct {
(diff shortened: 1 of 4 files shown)


radeksimko:
I'm not sure what's causing the failures in 1.14 and 1.15, but I'm pretty sure it's unrelated to this PR as I can reproduce it on `master`.

tfexec/errors.go
```diff
@@ -129,6 +129,16 @@ func (e *ErrVersionMismatch) Error() string {
 	return fmt.Sprintf("unexpected version %s (min: %s, max: %s)", e.Actual, e.MinInclusive, e.MaxExclusive)
 }
 
+func (e *ErrVersionMismatch) Is(target error) bool {
```
paultyng, line 132: I don't think we want this override, because you wouldn't want different version numbers to not match. I believe implementations are for `Is` to match errors of other types not of the same type with different data.

> An error type might provide an Is method so it can be treated as equivalent to an existing error. 

https://golang.org/pkg/errors/#Is

radeksimko: Ah, yes, that's a fair point, I'll remove it.

tfexec/errors.go
```diff
@@ -129,6 +129,11 @@ func (e *ErrVersionMismatch) Error() string {
...
+	_, ok := target.(*ErrVersionMismatch)
+	return ok
+}
+
```
paultyng, line 136: I think you can just remove this code entirely, the default implementation builtin to go should handle it.

```suggestion
```

paultyng (approved):

paultyng:
LGTM, but I think that `Is` function is unnecessary as its equivalent to the built-in code.

tfexec/errors.go
radeksimko: This is actually necessary because once it goes through `Is()`, the `errors` package handles unwrapping, which otherwise isn't handled and so `errors.Is` would return false on an unwrapped error.

PR #91, opened 2020-09-28
Enrich errors caused by context cancellation

radeksimko:
Previously the following `context.*` errors would be returned when the Terraform execution can't start because the context passed was already cancelled.

```
context deadline exceeded
```
```
context canceled
```

This is being replaced with more informative custom `ExitError`, e.g.

```
["/path/to/0.13.1/terraform" "version"] exited: context deadline exceeded
```

--- 

The following `exec.ExitError` error would be returned when execution is interrupted due to context (cancellation or deadline):

```
signal: killed
```

which is being replaced with `ExitError` too

```
["sleep" "1"] (pid 71492) exited (code -1): signal: killed
context deadline exceeded
```

--- 

Importantly the `ExitError` implements `Unwrap` to remain comparable with the original context errors or `exec.ExitError`, as demonstrated in attached tests.

--- 

We can make deadline errors even more informative by making timeout a first-class concept and implementing custom `context.Context` which reflects that. Once that is done, the error handling here should in theory just pick it up - only tests may need to be updated. I'll raise a separate PR for this though.

### DIFF
--- a/tfexec/apply.go
+++ b/tfexec/apply.go
@@ -91,7 +91,7 @@ func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
 	if err != nil {
 		return err
 	}
-	return tf.runTerraformCmd(cmd)
+	return tf.runTerraformCmd(ctx, cmd)
 }
 
 func (tf *Terraform) applyCmd(ctx context.Context, opts ...ApplyOption) (*exec.Cmd, error) {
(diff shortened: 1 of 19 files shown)


radeksimko:
Closing in favour of #94

radeksimko:
#94 does simplify error handling for consumers, but it doesn't actually address any of the above examples of errors.

I think there's still value in this PR, but also I understand if maintaining backwards compatibility for those who don't use `errors.Is/As` is important, especially if we still maintain Go 1.12 / 1.13 compatibility.

I still think that that _default_ error messaging should be improved and shouldn't be left to each consumer, but maybe it's a topic to revisit later, before we cut `v1` and have better justification to make such (potentially) breaking change.

PR #92, opened 2020-09-30
Cleanup tests

radeksimko:
I cleaned up tfexec tests:

 - we no longer _test_ for Go <1.14, so we can use `t.Cleanup()` which simplifies the test code
 - the global `tfVersion` seemed to do pretty much the same thing as the `testutil.TFCache`, i.e. cache installed versions of Terraform, so I just reused that which helped removing all global variables except for the one with cache instance

Let me know if I misread the original `tfVersion` implementation.


tfexec/internal/testutil/tfcache.go
```diff
@@ -79,7 +73,10 @@ func (tf *TFCache) find(t *testing.T, key string, finder func(dir string) tfinst
 		panic(fmt.Sprintf("unable to mkdir %q: %s", dir, err))
 	}
 
-	path, err := tfinstall.Find(context.Background(), finder(dir))
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	t.Cleanup(cancelFunc)
```
paultyng, line 77: there isn't a reason to use a cleanup here over a defer since its already function scoped, but I'm curious why use a cancelable context at all?

tfexec/terraform_test.go
```diff
@@ -5,30 +5,30 @@ import (
...
 
 	"github.com/hashicorp/terraform-exec/tfexec/internal/testutil"
-	"github.com/hashicorp/terraform-exec/tfinstall"
 )
 
+var tfCache *testutil.TFCache
+
 func TestMain(m *testing.M) {
 	os.Exit(func() int {
 		var err error
```
paultyng, line 17:
```suggestion
```

tfexec/internal/testutil/tfcache.go
radeksimko: defer != Cleanup, defer runs at the end of the function execution, i.e. when `find()` returns, whereas `t.Cleanup()` will run at the end of the test - e.g. after `t.Fatal()` or just when the test passes.

The cancellable context is useful when you want to just "Ctrl+C" running tests which are currently installing Terraform - presumably that would otherwise hang until the installation finishes.

kmoe (approved):
LGTM

tfexec/internal/testutil/tfcache.go
paultyng: I see so `t.Cleanup` runs when you press ctrl+c?

paultyng: This doesn't seem to happen for me in practice in a test project. I don't think cleanups are run when you Ctrl+C, at least on Windows.

paultyng: I can cancel the tests as well without waiting regardless of context.

paultyng: https://golang.org/pkg/os/signal/#hdr-Default_behavior_of_signals_in_Go_programs

SIGINT just causes it to exit unless a Notify has changed things from the default behavior.

radeksimko: I assume you're right (haven't tested it myself), but I still think this may be still useful for when the context is cancelled due to the timeout (i.e. exceeded `-timeout` as a flag given to `go test`), which doesn't involve any signals.

I'm going to merge it as is, but feel free to raise a separate issue or PR if you wish to discuss further.

Issue #73, opened 2020-09-04
Improve golden planfile test maintainability


kmoe:
The golden plan file used for Plan tests is specific to OS and Terraform version. Make this easier to maintain across development environments.

kmoe:
Some options for doing this:

1. Make a script that will regenerate the plan files for a given Terraform version and OS, and have devs run this whenever they bump a Terraform version.
  - Script could use Docker to run things under different OSs
2. Make a custom CI step, with manual trigger, which can do the same and push the new planfiles to the relevant PR branch.
3. Relax the check on plan files, so the `TerraformVersion` doesn't have to match for tests such as `TestShowPlanFile013` to pass. The test will still break if the plan file format changes in any other way, which is fine.

kmoe:
Step 3 is done, but this is not sufficient, because 0.13 is even stricter about planfile versions: in 0.13 `terraform show` itself will error (see https://github.com/hashicorp/terraform-exec/pull/93):

```
--- FAIL: TestShowPlanFile013 (1.56s)
    --- FAIL: TestShowPlanFile013/non_default_planfile_013-0.13.5 (1.56s)
        util_test.go:104: [INFO] running Terraform command: /tmp/tfinstall829262588/v-0.13.5/terraform init -no-color -force-copy -input=false -lock-timeout=0s -backend=true -get=true -get-plugins=true -lock=true -upgrade=false -verify-plugins=true
        util_test.go:104: [INFO] running Terraform command: /tmp/tfinstall829262588/v-0.13.5/terraform show -json -no-color planfilefoo
        show_test.go:311: 
            Error: Invalid plan file
            
            Failed to read plan from plan file: plan file was created by Terraform 0.13.0,
            but this is 0.13.5; plan files cannot be transferred between different
            Terraform versions.
```

PR #93, opened 2020-09-30
Bump to 0.13.3 for testing

paultyng:


paultyng:
The plan files need to be regnerated (or we need to tackle #73 )

kmoe (approved):
I've regenerated the 0.13 planfile and taken the opportunity to bump to 0.13.5.

As noted in https://github.com/hashicorp/terraform-exec/pull/102, 0.13 is stricter when it comes to planfiles. I suspect there is no way around the failure of `terraform show`, so a full solution to #102 is needed to do away with this inconvenience.

PR #94, opened 2020-09-30
Better support for errors.Is and errors.As

paultyng:
This potentially supersedes #91 

Probably best to review individual commits, the first commit splits exported errors to two files, one for exit error wrappers and one for the rest. No other changes beyond comments. The second commit adds the support for unwrapping and `errors.Is` comparisons.

### DIFF
--- a/tfexec/cmd_default.go
+++ b/tfexec/cmd_default.go
@@ -1,0 +1,45 @@
+// +build !linux
+
+package tfexec
+
+import (
+	"context"
+	"os/exec"
+	"strings"
+)
+
+func (tf *Terraform) runTerraformCmd(ctx context.Context, cmd *exec.Cmd) error {
+	var errBuf strings.Builder
+
+	cmd.Stdout = mergeWriters(cmd.Stdout, tf.stdout)
+	cmd.Stderr = mergeWriters(cmd.Stderr, tf.stderr, &errBuf)
+
+	go func() {
+		<-ctx.Done()
+		if ctx.Err() == context.DeadlineExceeded || ctx.Err() == context.Canceled {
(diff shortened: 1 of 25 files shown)


radeksimko (commented):
I'm ok with merging this (once test failures are addressed) - I suppose it's useful to have a single `errors.Is`/`errors.As` on the consumer side as opposed to two, but my main concern was lack of context in error messaging resulting from the default errors from stdlib.

I will address that differently in a different PR.

paultyng:
The test failure here is that interruption of `time_sleep` doesn't seem to work in 0.11, so I've just skipped < 0.12 for the timeout test.

paultyng:
Of course this succeeds in Windows which is where I wrote it 😭 I'll have to switch over to WSL and see whats up on linux.

tfexec/cmd_default.go
```diff
@@ -0,0 +1,43 @@
...
+
+	cmd.Stdout = mergeWriters(cmd.Stdout, tf.stdout)
+	cmd.Stderr = mergeWriters(cmd.Stderr, tf.stderr, &errBuf)
+
+	go func() {
+		<-ctx.Done()
+		if ctx.Err() == context.DeadlineExceeded || ctx.Err() == context.Canceled {
+			if cmd != nil && cmd.Process != nil && cmd.ProcessState != nil {
+				err := cmd.Process.Kill()
+				tf.logger.Printf("error from kill: %s", err)
```
kmoe, line 22: `if err != nil`?

paultyng: good catch!

paultyng:
@kmoe this is cleaned up now and passes with your plan file test fixes.

kmoe (approved):

PR #95, opened 2020-09-30
Provide ability to mock commands

radeksimko:
I plan to add some docs/comments to explain the usage and use cases, but otherwise the PR is ready for review and I consider it _functionally_ done.


paultyng:
This is going to be solved upstream in the LS

PR #98, opened 2020-10-06
Move gitref to sub package

paultyng:


tfinstall/gitref/git_ref_test.go
```diff
@@ -7,6 +7,9 @@ import (
 	"os/exec"
 	"strings"
 	"testing"
+
+	"github.com/hashicorp/terraform-exec/tfinstall"
+	"github.com/hashicorp/terraform-exec/tfinstall/gitref"
```
radeksimko, line 12: Once the package name is correct this import can be removed too.

tfinstall/gitref/git_ref.go
```diff
@@ -12,25 +12,27 @@ import (
...
 
-type GitRefOption struct {
+type Option struct {
 	installDir string
 	repoURL    string
 	ref        string
 }
 
-var _ ExecPathFinder = &GitRefOption{}
+var _ tfinstall.ExecPathFinder = &Option{}
```
radeksimko, line 25: Nitpick: If we move this to a test file then we can avoid the runtime dependency on `tfinstall`.

radeksimko (changes requested):

tfinstall/gitref/git_ref_test.go
```diff
@@ -1,4 +1,4 @@
-package tfinstall
+package gitref_test
```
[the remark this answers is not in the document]
paultyng: This is how you do only export testing.

radeksimko (approved):

PR #100, opened 2020-10-19
tfexec: Initial support for Terraform 0.14 -chdir global option

### LINKED ISSUE #99: Support Terraform 0.14.0 and later -chdir Global Option
Reference: https://github.com/hashicorp/terraform/blob/v0.14/CHANGELOG.md
Reference: https://github.com/hashicorp/terraform/pull/26087
Reference: https://github.com/hashicorp/terraform/issues/25558
Reference: https://github.com/hashicorp/terraform/issues/18030

Terraform commands have historical incompatibilities when executed outside the configuration directory, for example automatic variables files are not properly read. Terraform 0.14.0 (all pre-release versions) introduces a new `-chdir` global option, e.g. `terraform -chdir /path/to/config <command>` syntax. Later Terraform versions will deprecate the older `terraform <command> /path/to/config` as an argument.

bflad:
Closes #99

### DIFF
--- a/tfexec/providers_schema.go
+++ b/tfexec/providers_schema.go
@@ -2,17 +2,36 @@ package tfexec
 
 import (
 	"context"
+	"fmt"
 	"os/exec"
 
 	tfjson "github.com/hashicorp/terraform-json"
 )
 
+type providersSchemaConfig struct {
+	chdir string
+}
+
+var defaultProvidersSchemaOptions = providersSchemaConfig{}
+
+type ProvidersSchemaOption interface {
+	configureProvidersSchema(*providersSchemaConfig)
+}
+
+func (opt *ChdirOption) configureProvidersSchema(conf *providersSchemaConfig) {
+	conf.chdir = opt.path
+}
+
 // ProvidersSchema represents the terraform providers schema -json subcommand.
 func (tf *Terraform) ProvidersSchema(ctx context.Context) (*tfjson.ProviderSchemas, error) {
-	schemaCmd := tf.providersSchemaCmd(ctx)
+	schemaCmd, err := tf.providersSchemaCmd(ctx)
+
+	if err != nil {
+		return nil, err
+	}
 
 	var ret tfjson.ProviderSchemas
-	err := tf.runTerraformCmdJSON(ctx, schemaCmd, &ret)
+	err = tf.runTerraformCmdJSON(ctx, schemaCmd, &ret)
 	if err != nil {
 		return nil, err
 	}
@@ -25,9 +44,29 @@ func (tf *Terraform) ProvidersSchema(ctx context.Context) (*tfjson.ProviderSchem
 	return &ret, nil
 }
 
-func (tf *Terraform) providersSchemaCmd(ctx context.Context, args ...string) *exec.Cmd {
-	allArgs := []string{"providers", "schema", "-json", "-no-color"}
-	allArgs = append(allArgs, args...)
+func (tf *Terraform) providersSchemaCmd(ctx context.Context, opts ...ProvidersSchemaOption) (*exec.Cmd, error) {
+	c := defaultProvidersSchemaOptions
 
-	return tf.buildTerraformCmd(ctx, nil, allArgs...)
+	for _, o := range opts {
+		switch o.(type) {
+		case *ChdirOption:
+			err := tf.compatible(ctx, tf0_14_0, nil)
+			if err != nil {
+				return nil, fmt.Errorf("-chdir was added in Terraform 0.14: %w", err)
+			}
+		}
+
+		o.configureProvidersSchema(&c)
+	}
+
+	var args []string
+
+	// global opts
+	if c.chdir != "" {
+		args = append(args, "-chdir="+c.chdir)
+	}
+
+	args = append(args, []string{"providers", "schema", "-json", "-no-color"}...)
+
+	return tf.buildTerraformCmd(ctx, nil, args...), nil
 }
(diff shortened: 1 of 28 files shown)


paultyng:
Does `chdir` replace the work dir args that are in use in some plans already? We may need to look at this a bit more holistically, so we can utilize those options or something.

bflad:
@paultyng I believe so, please see https://github.com/hashicorp/terraform/pull/27664

bflad:
Rebased 👍  We will probably want to go through and mark all the removed "args" with `tf.compatible(ctx, nil, tf0_15_0)` as well -- happy to do that here or leave as a followup item.

tfexec/providers_schema.go
```diff
@@ -2,17 +2,36 @@ package tfexec
...
+type ProvidersSchemaOption interface {
+	configureProvidersSchema(*providersSchemaConfig)
+}
+
+func (opt *ChdirOption) configureProvidersSchema(conf *providersSchemaConfig) {
+	conf.chdir = opt.path
+}
+
 // ProvidersSchema represents the terraform providers schema -json subcommand.
 func (tf *Terraform) ProvidersSchema(ctx context.Context) (*tfjson.ProviderSchemas, error) {
```
radeksimko, line 26: I think in order to actually expose the `chdir` flag here we'd need to introduce `opts ...ProvidersSchemaOption` here as well, no?

tfexec/providers_schema_test.go
```diff
@@ -10,20 +10,44 @@ import (
 func TestProvidersSchemaCmd(t *testing.T) {
 	td := testTempDir(t)
 
-	tf, err := NewTerraform(td, tfVersion(t, testutil.Latest012))
+	tf, err := NewTerraform(td, tfVersion(t, testutil.Latest014))
```
radeksimko, line 13: I remember there were some format changes at some point - is it perhaps safer to test both 0.12 and >=0.14 which introduced the format change?

tfexec/workspace_list.go
```diff
@@ -24,6 +45,37 @@ func (tf *Terraform) WorkspaceList(ctx context.Context) ([]string, string, error
 	return ws, current, nil
 }
 
+func (tf *Terraform) workspaceListCmd(ctx context.Context, opts ...WorkspaceListCmdOption) (*exec.Cmd, error) {
+	// TODO: [DIR] param option
```
radeksimko, line 49: I think this is no longer relevant?
```suggestion
```

tfexec/workspace_list.go
```diff
@@ -3,18 +3,39 @@ package tfexec
...
+}
+
+func (opt *ChdirOption) configureWorkspaceList(conf *workspaceListConfig) {
+	conf.chdir = opt.path
+}
+
 // WorkspaceList represents the workspace list subcommand to the Terraform CLI.
-func (tf *Terraform) WorkspaceList(ctx context.Context) ([]string, string, error) {
+func (tf *Terraform) WorkspaceList(ctx context.Context, opts ...WorkspaceListCmdOption) ([]string, string, error) {
 	// TODO: [DIR] param option
```
radeksimko, line 28:
```suggestion
```
That doesn't seem relevant anymore

radeksimko (changes requested):
Sorry for the long delay in reviewing this @bflad - I finally found a spare moment to look at this PR of yours!

Do you have a use case (in SDK or elsewhere) for changing workdir during the lifetime of `*Terraform` instance?

I always worked with the assumption that the `*Terraform` is short-lived in a similar way to `context.Context`, which coincides with the fact that many commands didn't have an explicit argument (until recently) to run in a directory different to the one provided via `NewTerraform(workdir, ...)`.

I think it is fair to support both workflows, but I think it will require a little bit more effort involved to make this safe.

We currently persist `version` output in the struct
https://github.com/hashicorp/terraform-exec/blob/43ec6e17578bf2fd1e1037f9df5f1399d983eca8/tfexec/terraform.go#L56-L58
to be able to run compatibility checks without having to re-execute `version` repeatedly before each command.

The core version is (naturally) scoped to the Terraform binary (execPath), but the `version` command also outputs versions of providers which _are_ scoped to the workdir. I worry slightly that users won't realize this side effect while reading the cached output

https://github.com/hashicorp/terraform-exec/blob/43ec6e17578bf2fd1e1037f9df5f1399d983eca8/tfexec/version.go#L27-L36

Obviously this problem already exists given that we support the `dir` args in various commands already, but this is much more radical change which brings this problem more into spotlight IMO.

One way we could address this is by scoping the cache by workdir - as a map, or more likely as [`sync.Map`](https://pkg.go.dev/sync#Map) to make it safe for multiple commands to be executed at once. We should probably also consider some mechanism for purging the cache, like `Cleanup(workdir string)` - something that can be (optionally) called by downstream long-running process. This is to prevent any unnecessary memory consumption in case the consumer e.g. creates a high number of temporary directories in which it runs Terraform commands.

On a related note there is still a few commands which are missing the opts argument for passing `chdir`:
 - `ProvidersSchema`
 - `Version`
 - `Validate`

I think the `skipCache` argument should _probably_ stay there as 1st class argument (as opposed to being hidden under `opts ...VersionOption`) to communicate the side effect of the builtin cache.

> We will probably want to go through and mark all the removed "args" with tf.compatible(ctx, nil, tf0_15_0) as well -- happy to do that here

Yes please - if you can do that as part of this PR that would be great! 👍🏻

bflad:
Someone can recreate this, if they wish.

PR #102, opened 2020-11-06
Tests: relax golden planfile checks, better tfjson diffing, fix all other failures

kmoe:
This PR fixes all current test failures in master and offers a partial fix for #73, unblocking #93 .

Tests should no longer break when a new Core or provider version comes out, if there are no other changes to schema, statefiles, or planfiles.

 - [x] use go-cmp for better struct diffing in tests
 - [x] relax plan/statefile acctest checks: ignore changes in `TerraformVersion` attribute
 - [x] update null provider schema in `ProvidersSchema` acctest
 - [x] fix "Error: Could not satisfy plugin requirements" in `TestShowPlanFile012_linux` and `TestShowPlanFileRaw012_linux`


tfexec/internal/e2etest/providers_schema_test.go
```diff
@@ -136,8 +211,9 @@ func TestProvidersSchema(t *testing.T) {
 				}
 
 				expected := c.expected(tfv)
-				if !reflect.DeepEqual(schemas, expected) {
-					t.Fatalf("expected %+v, but got %+v", spew.Sdump(expected), spew.Sdump(schemas))
+
+				if diff := cmp.Diff(expected, schemas, cmp.Exporter(func(_ reflect.Type) bool { return true })); diff != "" {
```
radeksimko, line 215: I reckon the custom exporter is here because of cty? There is `ctydebug.CmpOptions` https://pkg.go.dev/github.com/zclconf/go-cty-debug@v0.0.0-20191215020915-b22d67c1ba0b/ctydebug#pkg-variables which provides equality comparator and decent output.

radeksimko (approved):

tfexec/internal/e2etest/providers_schema_test.go
kmoe: Yes, the issue here is that the cty types have a number of unexported fields, at paths too awkward to pass to go-cmp's `cmp.AllowUnexported` or `cmpopts.IgnoreUnexported`. Thanks for bringing `ctydebug` to my attention - I'll try that out.

tfexec/internal/e2etest/show_test.go
```diff
@@ -67,8 +66,8 @@ func TestShow(t *testing.T) {
 			t.Fatal(err)
 		}
 
-		if !reflect.DeepEqual(actual, &expected) {
-			t.Fatalf("actual: %s\nexpected: %s", spew.Sdump(actual), spew.Sdump(expected))
+		if diff := cmp.Diff(expected, actual, cmpopts.IgnoreFields(tfjson.State{}, "TerraformVersion")); diff != "" {
```
paultyng, line 69: you should wrap this in a `compareState` func or something to ensure all usages are the same

tfexec/internal/e2etest/providers_schema_test.go
```diff
@@ -25,88 +25,163 @@ func TestProvidersSchema(t *testing.T) {
...
 			"basic", func(tfv *version.Version) *tfjson.ProviderSchemas {
-				dk := tfjson.SchemaDescriptionKindPlain
-				// HACK: this is a bug? the contstant value is "plaintext" but "plain" is output,
-				// not sure if its in tfjson or what
-				dk = "plain"
-
-				providerName := "registry.terraform.io/hashicorp/null"
+				var providerSchema *tfjson.ProviderSchemas
 
 				if tfv.LessThan(version.Must(version.NewVersion("0.13.0"))) {
```
paultyng, line 30: You can reuse the global var `providerAddressMinVersion` here I think since this is really what you want to vary on.

paultyng: I wonder though if would be cleaner to just use a variable key for the map though? are there other parts that vary besides the key?

paultyng: Oh I see thats what it was already. Hard to see what the difference in the branches is here, if it was only those two values, why explode it out?

kmoe: There are several other differences between the two schemas, which made it messy to continue varying them inline. For example, the 0.13 schema has:
```
ConfigSchema: &tfjson.Schema{
  Version: 0,
  Block:   &tfjson.SchemaBlock{},
}
```

while the 0.12 schema has:

```
ConfigSchema: &tfjson.Schema{
  Version: 0,
    Block: &tfjson.SchemaBlock{
    DescriptionKind: "plain",
  },
}
```

Description types are different, and resource-level descriptions are only present from 0.13.

kmoe:
I'm wary of adding too many test helper functions - contributors should be able to edit individual tests and values within them easily without having to jump around to find out what is actually happening. The cleverer our test code is, the higher the risk that we are running incorrect tests. Tests should be stupid and verbose.

Nevertheless, tfjson comparison functions are convenient, so thanks for this, @paultyng.

PR #103, opened 2020-11-09
Bump go-getter to v1.5.0

picatz:
This PR introduces a version bump for go-getter to fix a dependency scanning warning found for this package.

github.com/ulikunitz/xz@0.5.5 used by github.com/hashicorp/go-getter@1.4.0 contains a DoS where it is possible to create an infinite read loop due to the usage of the ReadUvarint and ReadVarint function when encoding/binary via invalid inputs.

Related to https://github.com/hashicorp/go-getter/pull/278

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/google/go-cmp v0.5.2
 	github.com/hashicorp/go-checkpoint v0.5.0
 	github.com/hashicorp/go-cleanhttp v0.5.0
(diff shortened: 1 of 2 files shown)


kmoe:
Rebased for a clean test run. The go-getter update drops support for Go 1.12. We only use go-getter in tfinstall, which is being removed in https://github.com/hashicorp/terraform-exec/issues/109.

Prefer https://github.com/hashicorp/terraform-exec/issues/109 as a resolution to this, so closing this PR.

PR #105, opened 2020-11-17
Adds support for the terraform 0.12upgrade command

gdavison:
Adds support for the `terraform 0.12upgrade` command

### DIFF
--- a/tfexec/internal/e2etest/testdata/pre_011_syntax/file1.golden.txt
+++ b/tfexec/internal/e2etest/testdata/pre_011_syntax/file1.golden.txt
@@ -1,0 +1,31 @@
+resource "aws_instance" "web" {
+  ami           = data.aws_ami.amazon.id
+  instance_type = "t3.micro"
+  count         = 2
+
+  tags = {
+    Name = "HelloWorld"
+  }
+}
+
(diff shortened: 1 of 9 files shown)


radeksimko:
👋   Hey Graham,
What is the intended use case for this command?

In general `terraform-exec` is used for automation (i.e. usually unattended execution) and from what I can tell the upgrade commands are _not_ intended to be used in a "fire & forget" way, because there are no guarantees that the upgraded config will be actually usable and it may require further manual changes.

gdavison:
I'm using it for an enhancement to https://github.com/katbyte/terrafmt. It would allow users to apply `terraform 0.12` to any terraform blocks in Markdown or Go files.

kmoe (changes requested):
Thanks for this. The use case is valid and the implementation sound. Thanks for including a unit test.

Happy to merge when an end-to-end test is added. There's a section in [CONTRIBUTING.md](https://github.com/hashicorp/terraform-exec/blob/master/CONTRIBUTING.md#end-to-end-tests) describing how to do this. If you don't have time, please let me know and I'll be happy to add it.

kmoe (approved):
LGTM

PR #110, opened 2020-11-19
Support long terraform variable names in ErrMissingVar

c0sco:
An issue currently exists when a terraform variable with a very long name has no value. In these cases, the error message returned from tfexec/errors.go doesn't include the name of the terraform variable that is causing the problem. This is because terraform attempts to make error messages fit the terminal width, which means error messages can have a newline character instead of a space depending on the length of the variable name, and therefore the error message regex in terraform-exec doesn't match properly. This PR fixes this behavior to correctly display error messages no matter the length of the terraform variable name by matching on "\s" instead of " ".

This issue doesn't appear to be a problem with the Terraform 0.11 regex because of how much shorter the error message is.

To illustrate the problem, you can see the output when running the e2etest added in this PR without the fix to the regex in errors.go being in place:

> errors_test.go:74: expected missing no_default_really_long_variable_name_that_will_line_wrap_tf_output, got ""

### DIFF
--- a/tfexec/errors.go
+++ b/tfexec/errors.go
@@ -11,7 +11,7 @@ import (
 var (
 	// The "Required variable not set:" case is for 0.11
 	missingVarErrRegexp  = regexp.MustCompile(`Error: No value for required variable|Error: Required variable not set:`)
-	missingVarNameRegexp = regexp.MustCompile(`The root module input variable "(.+)" is not set, and has no default|Error: Required variable not set: (.+)`)
+	missingVarNameRegexp = regexp.MustCompile(`The root module input variable\s"(.+)"\sis\snot\sset,\sand\shas\sno\sdefault|Error: Required variable not set: (.+)`)
 
 	usageRegexp = regexp.MustCompile(`Too many command line arguments|^Usage: .*Options:.*|Error: Invalid -\d+ option`)
 
--- a/tfexec/internal/e2etest/errors_test.go
+++ b/tfexec/internal/e2etest/errors_test.go
@@ -43,7 +43,12 @@ func TestMissingVar(t *testing.T) {
 			t.Fatalf("err during init: %s", err)
 		}
 
-		_, err = tf.Plan(context.Background())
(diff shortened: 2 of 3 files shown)


tfexec/errors.go
```diff
@@ -11,7 +11,7 @@ import (
 var (
 	// The "Required variable not set:" case is for 0.11
 	missingVarErrRegexp  = regexp.MustCompile(`Error: No value for required variable|Error: Required variable not set:`)
-	missingVarNameRegexp = regexp.MustCompile(`The root module input variable "(.+)" is not set, and has no default|Error: Required variable not set: (.+)`)
+	missingVarNameRegexp = regexp.MustCompile(`The root module input variable\s"(.+)"\sis\snot\sset,\sand\shas\sno\sdefault|Error: Required variable not set: (.+)`)
```
radeksimko, line 14: Unless I'm mistaken, variables can't contain spaces, so would it make sense to instead just change `(.+)` to `([^ ]+)`? i.e. any character except for space? Would that satisfy the test and your needs?

c0sco: The issue is that any of the spaces after "The root module input variable" may be a newline. Here is an example of what I mean. You could have terraform output that looks like this:

> The root module input variable "a" is not set, and has no default

or this:

> The root module input variable "short_var" is not set, and has
> no default

or this:

> The root module input variable
> "super_long_variable_name_that_is_wrapped_to_its_own_line"
> is not set, and has no default

or it could look like this:

> The root module input variable "medium_name_var" is not set,
> and has no default

The line will wrap depending on the length of the variable name and what Terraform thinks the width of the terminal is. That can happen in multiple places in that error message (but always after the "The root module input variable" part of the string, which is why I'm not using \s before then).

radeksimko: Got it, that makes sense - thank you for all the examples.

radeksimko (approved):
LGTM, but I'll leave the decision with @kmoe as she has far more context here.

kmoe (approved):
Thanks, good fix!

PR #111, opened 2020-11-24
Update git ref test

kmoe:
As of https://github.com/hashicorp/terraform/commit/24d6c74e97430f4d0d78e879b487392fbf07914b, `master` on Core is now set to 0.15. Update tfinstall's git ref test to reflect this.

The `PR 25633` (https://github.com/hashicorp/terraform/pull/25633) test case started failing when the PR's branch was deleted 7 days ago. Since we can't predict when this will happen, I've removed the test case. I think we can be sure that if the other `refs/` cases work, PRs will too.

The commented-out test case `commit 83630a7` will be addressed in https://github.com/hashicorp/hcinstall/issues/1.


kmoe:
Some sort of juicy race condition going on here with the parallel tests. Git ref tests all work when run separately, so there's no serious issue at hand.

radeksimko (approved):

PR #113, opened 2020-12-16
Use json.Number instead of float64 when parsing state.

paddycarver:
State values can contain numbers that are larger than `float64` can
precisely represent. `json.Unmarshal` when unmarshaled into an
`interface{}` uses `float64` to represent numbers by default. To
properly retain precision for numbers, we need to use a custom
`json.Decoder` with the `UseNumber()` method called on it, which will
use `json.Number` (a `string` alias) to store the numbers, preserving
their precision.

This does this at the `runTerraformCmdJSON` level, meaning anyone using that method with `v` as `interface{}` in the arguments will suddenly find that their `float64`s are now `json.Number`s, which may break them. The solution to this is to use the `Float64()` method on the `json.Number`, but it is a code change they will need to make.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/hashicorp/go-getter v1.4.0
 	github.com/hashicorp/go-version v1.2.1
 	github.com/hashicorp/logutils v1.0.0
-	github.com/hashicorp/terraform-json v0.7.0
+	github.com/hashicorp/terraform-json v0.7.1-0.20201216193920-93fe74c2941e
 	github.com/mitchellh/cli v1.1.1
 	github.com/sergi/go-diff v1.1.0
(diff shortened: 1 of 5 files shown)


paultyng:
Can you put a test with like random or something in here that simulates this?

paddycarver:
I opened hashicorp/terraform-provider-corner#16, but I can see about doing something inline here, as well?

kmoe (approved):
LGTM. Test added in https://github.com/hashicorp/terraform-exec/pull/114, confirmed working with a version of the random provider compiled with https://github.com/hashicorp/terraform-plugin-sdk/commit/b78b8ce3dbf5cdadb2dfb4a2d7de831e5377fd29.

paultyng (approved):

PR #116, opened 2021-01-07
nightly test run; update test versions

### LINKED ISSUE #70: Run tests against Terraform Core master on CircleCI cron/scheduled job
Use `TFEXEC_E2ETEST_VERSIONS` as added in https://github.com/hashicorp/terraform-exec/pull/51/

kmoe:
fixes #70 

 - Update versions of 0.12 and 0.13 used in testing
 - Add 0.14 testing
 - Add nightly test run at midnight with failures reported to Slack
   - during the nightly run, a full test run is also carried out using the latest `refs/heads/master` of Terraform Core
   - the excessive configuration required to do this (a `post-step` and `context` on every job in the workflow) is a [known limitation](https://discuss.circleci.com/t/notify-the-slack-user-of-success-failure-of-an-entire-workflow-trigger-by-that-users-commit/34275/3) the CircleCI config syntax
 - Bonus config cleanup: add branch filter so pre-release jobs are not duplicated on PRs


radeksimko (approved):
Based on the related conversation on Slack - just to confirm - we don't intend to "stop supporting" 0.12 or any other earlier version explicitly, it's just that it will be supported on more best effort basis, am I right?

If so, then LGTM.

kmoe:
@radeksimko Yes, to confirm: 0.12 is still very much supported by terraform-exec, and acceptance tests for new and existing features are automatically run against the latest versions of 0.11, 0.12, 0.13, and 0.14. The only change this PR makes to 0.12 testing is that tests involving state file fixtures, in which the state file format is incompatible between pre-0.13 and 0.13+ versions, are now run for 0.13+ only.

Issue #109, opened 2020-11-19
Split out tfinstall into separate hcinstall module


kmoe:
https://github.com/hashicorp/hcinstall generalises tfinstall to all HashiCorp products with binaries on releases.hashicorp.com. It's in an experimental state with an API overhaul to improve usability.

Remember to include changes from https://github.com/hashicorp/terraform-exec/pull/111.

PR #119, opened 2021-01-12
0.15 beta testing

kmoe:
Commits will be appended to this PR and automatically tested as new 0.15 prerelease versions come out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,9 @@ commands:
       - run: go build ./...
   go_test:
     steps:
-      - run: go test -v ./...
+      - run:
+          no_output_timeout: 15m
+          command: go test -timeout=15m -v ./...
 
 jobs:
   # combined due to slowness of Go install
--- a/tfexec/internal/e2etest/util_test.go
+++ b/tfexec/internal/e2etest/util_test.go
@@ -33,6 +33,7 @@ func runTest(t *testing.T, fixtureName string, cb func(t *testing.T, tfVersion *
 		testutil.Latest012,
 		testutil.Latest013,
 		testutil.Latest014,
+		testutil.Latest015,
 	}
 	if override := os.Getenv("TFEXEC_E2ETEST_VERSIONS"); override != "" {
 		versions = strings.Split(override, ",")
--- a/tfexec/internal/testutil/tfcache.go
+++ b/tfexec/internal/testutil/tfcache.go
@@ -13,7 +13,7 @@ const (
 	Latest012 = "0.12.30"
 	Latest013 = "0.13.6"
 	Latest014 = "0.14.4"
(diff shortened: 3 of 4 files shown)


radeksimko:
(apologies for the force push, but I had to resolve some conflicts)

radeksimko:
I'm not sure why the tests are timing out on Windows:

> FAIL
> Too long with no output (exceeded 10m0s): context deadline exceeded

I will investigate.

.circleci/config.yml
```diff
@@ -26,7 +26,9 @@ commands:
       - run: go build ./...
   go_test:
     steps:
-      - run: go test -v ./...
+      - run:
+          no_output_timeout: 15m
+          command: go test -timeout=15m -v ./...
```
radeksimko, line 31: The `go test` timeout increase here is needed because it's a timeout for the whole package, not for a single test and the time it takes to run all tests within `e2etest` has (naturally) increased with another version in the mix. The default timeout is 10m and we would regularly reach ~8m without 0.15.

A cleaner solution here would be:
1. decoupling E2E tests entirely and make sure they never run by default on `go test` (without additional filtration)
2. create a step which runs just quick unit tests and that step can have default 10m timeout, or even lower
3. create a step which runs just E2E tests - here we can bump the timeout to whatever necessary

A followup on (3) could be also running these tests in parallel somehow in CircleCI, which could possibly help us bypass the per-package timeout? I'm not sure about that though - I don't have any experience with this.

.circleci/config.yml
```diff
@@ -26,7 +26,9 @@ commands:
       - run: go build ./...
   go_test:
     steps:
-      - run: go test -v ./...
+      - run:
+          no_output_timeout: 15m
```
radeksimko, line 30: I'm not entirely sure why we receive no output for more than `10m` and I don't see how to check how close we were to the default 10m timeout before, but this all seems to hang on `e2etest` package tests.

A cleaner solution here would be adding `t.Log` around operations which are likely to take long time - e.g. installation of Terraform, `init` (installation of providers) etc. but I avoided doing that for now, until the TF installation is decoupled per https://github.com/hashicorp/terraform-exec/issues/109

radeksimko (approved):
Just to make the bots happy I'm putting my 👍🏻  here (which should make the CI green), but I'd still appreciate @kmoe if you looked over the code and my comments before we merge it.

.circleci/config.yml
kmoe: I think this solution is fine, but have created https://github.com/hashicorp/terraform-exec/issues/145 in case we want to investigate (3).

kmoe:
Rebased following #146

radeksimko:
The last commit should also address the nightly failures.

PR #120, opened 2021-01-15
Handle terraform init flags removed in 0.15

kmoe:
From the Terraform 0.15 CHANGELOG:

* The `-lock` and `-lock-timeout` options are no longer available on `terraform init` [GH-27464]
* The `-verify-plugins=false` option is no longer available on `terraform init`. (Terraform now _always_ verifies plugins.) [GH-27461]
* The `-get-plugins=false` option is no longer available on `terraform init`. (Terraform now _always_ installs plugins.) [GH-27463]

This PR adds a compatibility error if the user attempts to set these options when using Terraform 0.15 and above. We maintain the existing behaviour of passing default values for these options prior to 0.15.

This fixes the errors seen in the nightly `master` test run ([example](https://app.circleci.com/pipelines/github/hashicorp/terraform-exec/623/workflows/d5b12efb-9755-4833-ac5e-08bf51f6b381/jobs/9136)).

### DIFF
--- a/tfexec/init.go
+++ b/tfexec/init.go
@@ -105,6 +105,14 @@ func (tf *Terraform) initCmd(ctx context.Context, opts ...InitOption) (*exec.Cmd
 	c := defaultInitOptions
 
 	for _, o := range opts {
+		switch o.(type) {
+		case *LockOption, *LockTimeoutOption, *VerifyPluginsOption, *GetPluginsOption:
+			err := tf.compatible(ctx, nil, tf0_15_0)
+			if err != nil {
+				return nil, fmt.Errorf("-lock, -lock-timeout, -verify-plugins, and -get-plugins options are no longer available as of Terraform 0.15: %w", err)
+			}
+		}
+
 		o.configureInit(&c)
 	}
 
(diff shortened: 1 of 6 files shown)


tfexec/init.go
```diff
@@ -114,17 +122,27 @@ func (tf *Terraform) initCmd(ctx context.Context, opts ...InitOption) (*exec.Cmd
...
 	// boolean opts: always pass
 	args = append(args, "-backend="+fmt.Sprint(c.backend))
 	args = append(args, "-get="+fmt.Sprint(c.get))
-	args = append(args, "-get-plugins="+fmt.Sprint(c.getPlugins))
-	args = append(args, "-lock="+fmt.Sprint(c.lock))
 	args = append(args, "-upgrade="+fmt.Sprint(c.upgrade))
-	args = append(args, "-verify-plugins="+fmt.Sprint(c.verifyPlugins))
+
+	// boolean opts removed in 0.15: pass if <0.15
+	err = tf.compatible(ctx, nil, tf0_15_0)
```
radeksimko, line 140: Nitpick: Should we also gate these flags on the other end (minimum version), similar to how we gate `fmt`?

radeksimko (approved):
Just one (not blocking) question/comment.

tfexec/init.go
kmoe: Yes, added https://github.com/hashicorp/terraform-exec/issues/121 as it's a somewhat bigger job to go through and find the minimum versions for all flags, which might as well be done all at once. Merging this PR now so we can see if the next nightly passes.

PR #122, opened 2021-01-20
Implemented state remove command

aleksanderaleksic:
Implemented the state remove command, with a lot of inspiration from the state move command

### DIFF
--- a/tfexec/state_rm.go
+++ b/tfexec/state_rm.go
@@ -1,0 +1,104 @@
+package tfexec
+
+import (
+	"context"
+	"os/exec"
+	"strconv"
+)
+
+type stateRmConfig struct {
+	backup      string
+	backupOut   string
+	dryRun      bool
+	lock        bool
+	lockTimeout string
+	state       string
+	stateOut    string
+}
+
+var defaultStateRmOptions = stateRmConfig{
+	lock:        true,
+	lockTimeout: "0s",
+}
+
+// StateRmCmdOption represents options used in the Refresh method.
+type StateRmCmdOption interface {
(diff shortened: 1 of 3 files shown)


paultyng:
@aleksanderaleksic this mostly looks good to me, do you think you could also add an e2e test (see the `tfexec/internal/e2etest` directory). That will at least let it get exercised across multiple versions of TF to ensure its present and works in an expected way.

aleksanderaleksic:
@paultyng Added a test now 👍🏻

paultyng (approved):
LGTM, would like to have @kmoe give a once over as well though.

tfexec/state_rm.go
```diff
@@ -0,0 +1,104 @@
...
+func (opt *StateOption) configureStateRm(conf *stateRmConfig) {
+	conf.state = opt.path
+}
+
+func (opt *StateOutOption) configureStateRm(conf *stateRmConfig) {
+	conf.stateOut = opt.path
+}
+
+// StateRm represents the terraform state rm subcommand.
+func (tf *Terraform) StateRm(ctx context.Context, source string, opts ...StateRmCmdOption) error {
```
kmoe, line 58: `address` would be a clearer name for this parameter than `source` here, I think, corresponding to the cmd help text:

```
$ terraform -help state rm
Usage: terraform state rm [options] ADDRESS...
```

kmoe (approved):
Thank you, looks good. One comment regarding variable naming.

tfexec/state_rm.go
aleksanderaleksic: Yep, I agree will fix 👍🏻

PR #124, opened 2021-01-28
Notify on nightly success as well as failure

kmoe:
Also:
-  Remove Slack notification step from Windows build jobs, because it doesn't work with the Windows build agent (known issue with CircleCI orb: https://github.com/CircleCI-Public/slack-orb/issues/208, https://github.com/CircleCI-Public/slack-orb/issues/217).
- Remove use of CircleCI contexts to import environment variables, due to change by Release Engineering.

The success notification has been tested.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -252,12 +252,7 @@ workflows:
               only:
                 - master
     jobs:
-      - winbuildtest:
-          context: terraform-exec-slack-notifications
-          post-steps:
-            - slack/notify:
-                event: fail
-                template: basic_fail_1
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
I worry slightly that this would lead to people leaving the #proj-terraform-exec channel as they might have joined to watch and discuss dev questions, but may not care as much about test results.

Have you considered sending these to a separate `feed` channel?

kmoe:
Thanks @radeksimko, that's a great idea. The channel is set through an env var, so I can update this in CircleCI later after creating the channel.

PR #127, opened 2021-02-23
update null provider schema in tests

kmoe:
Version 3.1.0 of the null provider, released last Friday, changes some fields in the schema, which we read in the `TestProvidersSchema` e2e test. This PR updates them and fixes the provider version in the test config file.

This PR fixes the current nightly test failure.


paddycarver (approved):
Looks good to me. Was the schema update on the null provider not a breaking change?

kmoe:
All that's changed in the schema are the descriptions, (soft) deprecation of one attribute, and `Optional` set to `false` for `id`, which I believe does not change functionality for any end-user. It looks like providers might be another example of semver being a bit ambiguous.

PR #130, opened 2021-03-04
Update readme example

vfiftyfive:
The current syntax in the README doesn't work.
`tfinstall.Find` is missing one argument in the function call.
This PR adds `context.Background()` as an additional parameter.

Nic


radeksimko (approved):
Good catch! Thank you for the PR.

PR #138, opened 2021-03-10
Parse version -json output and fall back to plaintext

### LINKED ISSUE #13: Parse terraform version -json output
From 0.13 onwards, `terraform version` accepts a `-json` flag. See https://github.com/hashicorp/terraform/pull/25252.

It might be worth trying `terraform version -json` first, before falling back to the legacy version output parsing logic added in #7.

radeksimko:
Closes #13 

I didn't add more tests because it seemed like we'd be just testing mostly stdlib internals - either os.Exec or encoding/json. I'm open to ideas on what else to test though.

Also I think that exposing `VersionOutput` from the `Version()` method would be a good idea, but I avoided that just to retain compatibility, for now. Also it might be helpful to leverage (not yet published) library for parsing the provider addresses too. Either way I can open a separate issue for this.

### DIFF
--- a/tfexec/version.go
+++ b/tfexec/version.go
@@ -3,12 +3,14 @@ package tfexec
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/terraform-json"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


kmoe (approved):
This worked out better than I expected, as it looks like earlier Terraform versions don't error when `-json` is supplied to `terraform version` - they just ignore it. 

Agree with testing approach here. We do need to add e2e tests for the `providerVersions` (see TODO in `e2etest/version_test.go`, but this can be done here or elsewhere. Let's create an issue if this isn't done now.

What would be the use case for exposing `VersionOutput`?

radeksimko:
> What would be the use case for exposing VersionOutput?

I don't have a strong use case for it actually, I just thought maybe in some contexts it would be useful to know the platform, but I guess if it already differs from the host platform then it's unlikely we'd even be able to execute terraform anyway, so I guess there isn't a strong reason 🤷🏻

kmoe:
Great, let's keep a lookout for use cases until we're ready for 1.0, because now is the time to break things if we want to!

radeksimko:
Created https://github.com/hashicorp/terraform-exec/issues/147

PR #146, opened 2021-03-29
Fix validate test in 0.15 prereleases

kmoe:
As discussed elsewhere, versions such as `0.15.0-beta2` and `0.15.0-dev` do not satisfy the condition `tfv.GreaterThanOrEqual(version.Must(version.NewVersion("0.15.0")))`. A proposal will be made separately to `go-version` to provide prerelease-compatible comparison functions. In the meantime, hardcode the beta and nightly version numbers into the test so both tests pass.


radeksimko (approved):
LGTM, as long as the CI is green.

PR #159, opened 2021-04-08
add refs/heads/main test on PR and release builds

kmoe:
To rebase once the current fail on `main` is addresed.


radeksimko (approved):
This LGTM, although my confusion came mostly from the fact that this isn't default for when I ran `go test` locally, without any ENV variables.

kmoe:
I'd prefer not to have the `refs/heads/main` test run by default during local development for now, as it takes significantly more time and CPU.

PR #160, opened 2021-04-08
Fix E2E test failures caused by misplaced Core()

radeksimko:


kmoe (approved):
Tested `refs/heads/main` locally

PR #161, opened 2021-04-13
circle: Run brew update before installing go

radeksimko:
This is to address installation failure observed last night: https://app.circleci.com/pipelines/github/hashicorp/terraform-exec/855/workflows/4f4a773f-e4d4-4991-8a0a-a075560601d1/jobs/11437

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,12 +14,13 @@ commands:
   macos_install_go:
     steps:
       - run:
-          # just latest stable version of Go as a sanity check
-          # version specification would require a brew update probably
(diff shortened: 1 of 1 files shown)


radeksimko:
Actually I wonder if just bumping XCode version would give us more up-to-date Homebrew installation pre-installed
https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions

Let me try that...

radeksimko:
ok, that didn't help, so `brew update` is the best solution for now 🤷🏻‍♂️

kmoe (approved):

PR #162, opened 2021-04-14
Create a constant for the binary file name

sundowndev:
As a user, I'd like to predicate the binary file name tfexec will try to execute. This can be useful when you need to install multiple versions of Terraform at the same time in different directories but install each version only once. If that binary file name might change in the future, I don't want my code to fail because of that. Is that something relevant to you? May be I missed a method to achieve that differently, let me know!

### References

- https://github.com/cloudskiff/driftctl/pull/416

### DIFF
--- a/tfinstall/download.go
+++ b/tfinstall/download.go
@@ -92,7 +92,7 @@ func downloadWithVerification(ctx context.Context, tfVersion string, installDir 
 		return "", err
 	}
 
-	return filepath.Join(tfDir, "terraform"), nil
+	return filepath.Join(tfDir, BinaryFileName), nil
 }
 
 // verifySumsSignature downloads SHA256SUMS and SHA256SUMS.sig and verifies
--- a/tfinstall/look_path.go
+++ b/tfinstall/look_path.go
@@ -18,7 +18,7 @@ func LookPath() *LookPathOption {
 }
 
 func (opt *LookPathOption) ExecPath(context.Context) (string, error) {
-	p, err := exec.LookPath("terraform")
+	p, err := exec.LookPath(BinaryFileName)
 	if err != nil {
 		if notFoundErr, ok := err.(*exec.Error); ok && notFoundErr.Err == exec.ErrNotFound {
 			log.Printf("[WARN] could not locate a terraform executable on system path; continuing")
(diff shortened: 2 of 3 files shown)


radeksimko:
Hi @sundowndev 
As described in #109, the `tfinstall` package will be removed in the next release. We recommend that https://github.com/hashicorp/hc-install be used instead. Let us know what you think of the API.

It does expose the binary name for [two products](https://pkg.go.dev/github.com/hashicorp/hc-install@v0.1.0/product) currently (including Terraform), but I would encourage you to perhaps look at the `Remove` method of the installer instead, which can do the cleanup for you. That way you don't have to deal with any particular binary name in your application and treat it as an implementation detail.

https://pkg.go.dev/github.com/hashicorp/hc-install#Installer.Remove

Here is an example snippet which demonstrates how you can install multiple versions:

```go
package main

import (
	"context"
	"log"
	"os"

	"github.com/hashicorp/go-version"
	"github.com/hashicorp/hc-install"
	"github.com/hashicorp/hc-install/product"
	"github.com/hashicorp/hc-install/releases"
	"github.com/hashicorp/hc-install/src"
)

func main() {
	tmpDir := os.TempDir()
	err := run(tmpDir)
	if err != nil {
		log.Fatal(err)
	}
}

func run(workDir string) error {
	ctx := context.Background()

	cons, _ := version.NewConstraint(">= 0.12")
	rv := &releases.Versions{
		Product:     product.Terraform,
		Constraints: cons,
	}

	versions, err := rv.List(ctx)
	if err != nil {
		return err
	}

	i := install.NewInstaller()
	for _, installableVersion := range versions {
		ver := installableVersion.(*releases.ExactVersion)
		log.Printf("installing %s...", ver.Version)

		execPath, err := i.Ensure(context.Background(), []src.Source{
			installableVersion,
		})
		if err != nil {
			return err
		}

		log.Printf("version %s installed", ver.Version)

		// Do some testing here
		if err != nil {
			return err
		}
		tfVersion, _, err := tf.Version(ctx, true)
		if err != nil {
			return err
		}
		log.Printf("TF Version: %q", tfVersion)
		// tf.Destroy(ctx)
		// etc.
	}
	i.Remove(ctx)

	return nil
}
```

Just keep in mind that the installation and removal are not (yet) thread-safe, so each installation should happen in a single thread and removal afterwards. See also https://github.com/hashicorp/hc-install/issues/35 for more on that topic.

PR #166, opened 2021-04-22
security: update hashicorp pgp key

kmoe:
HashiCorp security bulletin: [HCSEC-2021-12](https://discuss.hashicorp.com/t/hcsec-2021-12-codecov-security-event-and-hashicorp-gpg-key-exposure/23512).

The HashiCorp public key used for signing `terraform` releases has been rotated. 

All `terraform` releases have been re-signed with the new key with ID `72D7468F`. The `tfinstall` package now verifies signature files generated using this new key, which have the suffix `SHA256SUMS.72D7468F.sig`.


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+# 0.13.3 (Unreleased)
+
+SECURITY:
+ - `tfinstall`: The HashiCorp PGP signing key has been rotated (HCSEC-2021-12). This key is used to verify downloaded versions of Terraform. We recommend all users of terraform-exec upgrade to v0.13.3 for this security fix. [GH-166]
```
radeksimko, line 4: Shall we also link to the bulletin from here and add something like

> Older versions of Terraform are temporarily retaining the old (now revoked) key, but that key will be removed. Installation via older `terraform-exec` versions will therefore stop working soon and you should upgrade as soon as possible to avoid any interruption.

tfinstall/download.go
```diff
@@ -65,7 +65,7 @@ func downloadWithVerification(ctx context.Context, tfVersion string, installDir
 	client.Dst = sumsTmpDir
 	err = client.Get()
 	if err != nil {
-		return "", fmt.Errorf("error fetching checksums: %s", err)
+		return "", fmt.Errorf("error fetching checksums at URL %s: %s", sumsURL, err)
```
radeksimko, line 68:
```suggestion
		return "", fmt.Errorf("error fetching checksums at URL %s: %w", sumsURL, err)
```

radeksimko (approved):

CHANGELOG.md
kmoe: Excellent idea, thank you. Added an extra note and the link.

radeksimko (approved):

PR #167, opened 2021-04-23
tfexec/testutil: Bump latest stable versions

radeksimko:
The tests will be failing until new patch releases are out, as per https://discuss.hashicorp.com/t/hcsec-2021-12-codecov-security-event-and-hashicorp-gpg-key-exposure/23512


kmoe:
Rerunning workflow following release of patched Terraform versions.

kmoe (approved):
Thanks for this and for adding the 0.14 statefile test

PR #168, opened 2021-04-23
Update GPG usage

shore:
As part of rotating the GPG key used for signing products, we are also moving the GPG bits into a dedicated CircleCI context.  This PR renames the GPG env vars to match those from the new context and applies the context to the job that uses the key.


shore:
Rotating this key is not needed.

PR #170, opened 2021-04-29
Add method for listing available versions

mwhooker:
Adds `ListVersions` which returns a sorted list of terraform versions. 

My use-case is wanting to write a test matrix which tests against every version of tf within a range.


radeksimko:
Hey @mwhooker 
Sorry for the delay in reviewing your PR.

As you may or may not be aware, we were working on `hc-install` which will became the preferred way of installing Terraform (and also other HashiCorp tools) from the upcoming tfexec version, which will have `tfinstall` removed.

--- 

I'm attaching a snippet that can provide the same functionality using hc-install API:

```go
package main

import (
	"context"
	"log"
	"os"

	"github.com/hashicorp/go-version"
	"github.com/hashicorp/hc-install"
	"github.com/hashicorp/hc-install/product"
	"github.com/hashicorp/hc-install/releases"
	"github.com/hashicorp/hc-install/src"
)

func main() {
	tmpDir := os.TempDir()
	err := run(tmpDir)
	if err != nil {
		log.Fatal(err)
	}
}

func run(workDir string) error {
	ctx := context.Background()

	cons, _ := version.NewConstraint(">= 0.12")
	rv := &releases.Versions{
		Product:     product.Terraform,
		Constraints: cons,
	}

	versions, err := rv.List(ctx)
	if err != nil {
		return err
	}

	i := install.NewInstaller()
	for _, installableVersion := range versions {
		ver := installableVersion.(*releases.ExactVersion)
		log.Printf("installing %s...", ver.Version)

		execPath, err := i.Ensure(context.Background(), []src.Source{
			installableVersion,
		})
		if err != nil {
			return err
		}

		log.Printf("version %s installed", ver.Version)

		// Do some testing here
		if err != nil {
			return err
		}
		tfVersion, _, err := tf.Version(ctx, true)
		if err != nil {
			return err
		}
		log.Printf("TF Version: %q", tfVersion)
		// tf.Destroy(ctx)
		// etc.
	}
	i.Remove(ctx)

	return nil
}
```

Let us know over in `hc-install` issue tracker if this doesn't work for you - we can further tweak the API and discuss your requirements.

Thanks!

PR #175, opened 2021-05-10
Added documentation on exact_version.go

plallin:
I find `tfinstall` a bit underdocumented. None of its methods or function has comments, cf [its official documentation](https://pkg.go.dev/github.com/hashicorp/terraform-exec/tfinstall).

I needed to understand the role and behavior of some function / method, so I added documentation.

Comments welcome.

### DIFF
--- a/tfinstall/exact_version.go
+++ b/tfinstall/exact_version.go
@@ -15,6 +15,9 @@ type ExactVersionOption struct {
 
 var _ ExecPathFinder = &ExactVersionOption{}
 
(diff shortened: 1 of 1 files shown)


kmoe:
Thank you for adding this documentation. I agree `tfinstall` was underdocumented. As an API experiment it has served its purpose and will be removed in the next release (see #109). We recommend you use https://github.com/hashicorp/hc-install instead - let us know what you think of the API.

PR #187, opened 2021-06-07
fix: find "terraform" command if execPath is empty in NewTerraform

suzuki-shunsuke:
#188 

This pull request changes NewTerraform to find `terraform` command by exec.LookPath if execPath is empty.

### DIFF
--- a/tfexec/terraform.go
+++ b/tfexec/terraform.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"os/exec"
 	"sync"
 
 	"github.com/hashicorp/go-version"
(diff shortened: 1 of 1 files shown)


radeksimko (commented):
Hi @suzuki-shunsuke 
I'm sorry for the delay in reviewing your PR.

I don't want to make a final decision on this _yet_ before discussing with the rest of the team, but I wanted to share my thoughts here as some food for thought.

I understand your pragmatic intentions here but I find the existing design (where we error out on empty path) actually helpful in forcing (especially new) user to think about the consequences they probably wouldn't have thought about if we did auto-lookup like you suggest.

There are differences between major Terraform versions and the same person managing the app consuming `terraform-exec` may not be the same person managing Terraform installation e.g. in CI. For example GitHub manages default Terraform installation in the default GitHub Actions image and downstream users of GHA which may utilize `terraform-exec` have no control over what version of Terraform ends up being installed there and when/whether it does get upgraded.

Forcing the user to run `exec.LookPath("terraform")` explicitly and passing that path to the library as a way of acknowledging that _"I'm okay with any discovered version and I know the risk"_, or **better** _"I manage the version and I always know what to expect"_.

suzuki-shunsuke:
Hi @radeksimko 
Thank you for your review.

I understand your thought.
If we keep the current behavior, how about updating the description of `NewTerraform`?

https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.14.0/tfexec#NewTerraform

> If a blank execPath is supplied, NewTerraform will attempt to locate an appropriate binary on the system PATH.

This description isn't correct.

kmoe:
Thank you for your PR, @suzuki-shunsuke. As described in https://github.com/hashicorp/terraform-exec/issues/109, the `tfinstall` will be removed in the next release. We recommend that https://github.com/hashicorp/hc-install be used instead. Let us know what you think of the API - it should be easier to specify an installer that falls back to `exec.LookPath`.

PR #195, opened 2021-06-23
Use Go 1.16 for building Terraform

radeksimko:
Terraform recently introduced some code which requires Go 1.16
https://github.com/hashicorp/terraform/pull/28874
as can be seen from our failing acceptance tests:

```
        # github.com/hashicorp/terraform/internal/command/views
        internal/command/views/add.go:76:10: undefined: os.WriteFile
```

https://golang.org/pkg/os/#WriteFile

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,11 +25,18 @@ commands:
   go_build:
     steps:
       - run: go build ./...
-  go_test:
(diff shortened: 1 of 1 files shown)


radeksimko:
Merging to get E2E nightly test runs green again, but reviews are still welcomed.

PR #199, opened 2021-07-03
(maint) update scripts/release/release.sh

vladdoster:
- use  `$(...)` over  ``` `...` ``` (backticks) because it is the legacy syntax required by only the very oldest of non-POSIX-compatible bourne-shells
- remove un-used `START_DIR` variable
- use `printf` bringing it up to date with the rest of the script and honors the `\n` without needing a `-e` flag.

### DIFF
--- a/scripts/release/release.sh
+++ b/scripts/release/release.sh
@@ -18,8 +18,7 @@ function pleaseUseGNUsed {
 function init {
   sed --version > /dev/null || pleaseUseGNUsed
 
-  DATE=`date '+%B %d, %Y'`
-  START_DIR=`pwd`
+  DATE=$(date '+%B %d, %Y')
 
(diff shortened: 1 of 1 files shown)


vladdoster:
@jptoto @aareet

Does this repo take external PRs?

radeksimko (commented):
Hi @vladdoster 
Sorry for the delay in reviewing your PR.

The changes you propose look reasonable to me but may I ask you what motivated the PR since all these changes don't seem to fix any particular bug - unless I'm mistaken?

I should mention that we use pretty much the same script in a number of other repositories: https://github.com/search?q=org%3Ahashicorp+getTargetVersion&type=code

radeksimko (approved):

PR #206, opened 2021-08-10
Add StateList

micahkemp:
This PR adds the StateList command.

### DIFF
--- a/tfexec/options.go
+++ b/tfexec/options.go
@@ -4,6 +4,16 @@ import (
 	"encoding/json"
 )
 
+// AddressOption represents an address.
+type AddressOption struct {
+	address string
+}
+
+// Address represents an address.
+func Address(address string) *AddressOption {
+	return &AddressOption{address}
+}
+
 // AllowMissingConfigOption represents the -allow-missing-config flag.
 type AllowMissingConfigOption struct {
 	allowMissingConfig bool
@@ -164,6 +174,16 @@ type GetPluginsOption struct {
 
 func GetPlugins(getPlugins bool) *GetPluginsOption {
 	return &GetPluginsOption{getPlugins}
+}
+
+// IdOption represents the -id flag.
+type IdOption struct {
+	id string
+}
+
+// Id represents the -id flag.
+func Id(id string) *IdOption {
+	return &IdOption{id}
 }
 
 // LockOption represents the -lock flag.
--- a/tfexec/internal/e2etest/state_list_test.go
+++ b/tfexec/internal/e2etest/state_list_test.go
@@ -1,0 +1,35 @@
+package e2etest
+
+import (
+	"context"
+	"reflect"
+	"testing"
+
+	"github.com/hashicorp/go-version"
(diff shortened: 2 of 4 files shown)


tfexec/options.go
```diff
@@ -4,6 +4,16 @@ import (
 	"encoding/json"
 )
 
+// AddressOption represents an address.
```
micahkemp, line 7: There may be opinions regarding how this is named, or even implemented. `terraform state show` is documented as taking zero or more addresses as options.  As such I added an AddressOption, which ends up getting applied (for `StateList`) as positional arguments after all the other options.

I'm more than happy to change how this is implemented if there are other suggestions.

radeksimko (commented):
Thank you for this PR as well.

I have the same question for this command as in your [other PR](https://github.com/hashicorp/terraform-exec/pull/205):
Could you share your use case for this command in the context of automation?

Keep in mind that this command (unlike most others we integrated) does not provide machine-readable (JSON) output and AFAIK therefore doesn't provide any compatibility guarantees for the output format.

If we have a use case though we could propose enhancement of Terraform CLI so that it does offer JSON output - this would then make the integration much easier here.

micahkemp:
Use case described [here](https://github.com/hashicorp/terraform-exec/pull/205#issuecomment-896089101) on #205.

I'll also mention that as these two PRs are related in terms of how I would use them, and because they're both being submitted at the same time, I'm more than happy to combine them into one.  I kept them split to hopefully make them simpler to review and also to allow you flexibility in reviewing them as separate concepts.

micahkemp:
As discussed in the comments of #205, I think it makes sense for wait until `state show` supports `-json`.

radeksimko:
Yep, I will be doing some exploratory work which will likely result adding JSON output for more commands and I will certainly use your use case in the process to advocate for `state list -json` specifically.

I will let you know here once any relevant PR is ready upstream.

Thanks.

radeksimko:
As mentioned in https://github.com/hashicorp/terraform-exec/pull/205#issuecomment-1172203797 the use case can also be satisfied with existing functionality. See below.

```sh
$ terraform state list
random_pet.adam
random_pet.basil
random_pet.claire
```
which is equivalent to
```sh
$ terraform show -json | jq -r '.values.root_module.resources[].address'
random_pet.adam
random_pet.basil
random_pet.claire
```
and similarly to
```go
package main

import (
	"context"
	"fmt"
	"log"
	"os"
	"os/exec"

)

func main() {
	workDir, err := os.Getwd()
	if err != nil {
		log.Fatal(err)
	}

	execPath, err := exec.LookPath("terraform")
	if err != nil {
		log.Fatal(err)
	}

	if err != nil {
		log.Fatal(err)
	}

	ctx := context.Background()
	state, err := tf.Show(ctx)
	if err != nil {
		log.Fatal(err)
	}

	for _, resource := range state.Values.RootModule.Resources {
		fmt.Println(resource.Address)
	}
}
```

I hope that helps.

PR #209, opened 2021-08-12
add supports for cmd: terraform add

### LINKED ISSUE #208: Add supports for `terraform add` command
The v1.1.0 of `terraform` introduces a new subcommand: `terrraform add`.

It would be very helpful to have it available in `terraform-exec`, which can be used for implementing some "bulk" import feature.

magodo:
Fix #208 

Though the official release of the v1.1.0 is not shipped yet, while the alpha version is available. This means general users can use the `terraform add` feature already via the alpha CLI tool. As the `tfinstall` is able to install the alpha release, it means the `tfexec` users should also be able to use the alpha features (i.e. the `add`).

For the reason above, I've added a `Latest_v1_1_alpha` version for the test.

### DIFF
--- a/tfexec/add.go
+++ b/tfexec/add.go
@@ -1,6 +1,7 @@
 package tfexec
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os/exec"
@@ -41,13 +42,25 @@ func (opt *ReattachOption) configureAdd(conf *addConfig) {
 	conf.reattachInfo = opt.info
 }
 
-// Add represents the terraform add subcommand.
-func (tf *Terraform) Add(ctx context.Context, address string, opts ...AddOption) error {
+// Add represents the `terraform add` subcommand (added in 1.1.0).
+//
+// Note that this function signature and behaviour is subject
+// to breaking changes including removal of that function
+// until final 1.1.0 Terraform version (with this command) is released.
+func (tf *Terraform) Add(ctx context.Context, address string, opts ...AddOption) (string, error) {
 	cmd, err := tf.addCmd(ctx, address, opts...)
 	if err != nil {
-		return err
+		return "", err
 	}
-	return tf.runTerraformCmd(ctx, cmd)
+
+	var outBuf bytes.Buffer
+	cmd.Stdout = mergeWriters(cmd.Stdout, &outBuf)
+
+	if err := tf.runTerraformCmd(ctx, cmd); err != nil {
+		return "", err
+	}
+
+	return outBuf.String(), nil
 }
 
 func (tf *Terraform) addCmd(ctx context.Context, address string, opts ...AddOption) (*exec.Cmd, error) {
--- a/tfexec/add.go
+++ b/tfexec/add.go
@@ -1,0 +1,88 @@
+package tfexec
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strconv"
+)
+
+type addConfig struct {
+	fromState       bool
+	out             string
+	includeOptional bool
+	provider        string
+	reattachInfo    ReattachInfo
+}
+
+var defaultAddOptions = addConfig{}
+
+type AddOption interface {
+	configureAdd(*addConfig)
+}
+
+func (opt *FromStateOption) configureAdd(conf *addConfig) {
+	conf.fromState = opt.fromState
+}
+
+func (opt *OutOption) configureAdd(conf *addConfig) {
+	conf.out = opt.path
+}
+
+func (opt *IncludeOptionalOption) configureAdd(conf *addConfig) {
+	conf.includeOptional = opt.includeOptional
+}
+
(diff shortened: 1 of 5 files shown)


tfexec/internal/testutil/tfcache.go
```diff
@@ -9,12 +9,13 @@ import (
...
-	Latest014 = "0.14.11"
-	Latest015 = "0.15.5"
-	Latest_v1 = "1.0.0"
+	Latest011         = "0.11.15"
+	Latest012         = "0.12.31"
+	Latest013         = "0.13.7"
+	Latest014         = "0.14.11"
+	Latest015         = "0.15.5"
+	Latest_v1         = "1.0.0"
+	Latest_v1_1_alpha = "1.1.0-alpha20210811"
```
radeksimko, line 18:
```suggestion
	Latest_v1_1 = "1.1.0-alpha20210811"
```

tfexec/add.go
```diff
@@ -0,0 +1,88 @@
...
+func (opt *ProviderOption) configureAdd(conf *addConfig) {
+	conf.provider = opt.provider
+}
+
+func (opt *ReattachOption) configureAdd(conf *addConfig) {
+	conf.reattachInfo = opt.info
+}
+
+// Add represents the terraform add subcommand.
+func (tf *Terraform) Add(ctx context.Context, address string, opts ...AddOption) error {
```
radeksimko, line 45:
```suggestion
func (tf *Terraform) Add(ctx context.Context, address string, opts ...AddOption) (string, error) {
```

The `add` command writes configuration to stdout by default and so I think that the interface of this method should reflect that default behaviour and probably just entirely omit support for the `out` flag.

IMHO in contexts where `terraform-exec` is used users/developers can still process the string and write it to a file if they choose to.

tfexec/internal/testutil/tfcache.go
radeksimko: Technically this is an internal constant anyway, so it probably doesn't matter if we later rename it but I expect that we'd more likely be testing against any latest v1.1 once a stable version is out rather than keep testing against alpha versions.

tfexec/add.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (opt *ProviderOption) configureAdd(conf *addConfig) {
+	conf.provider = opt.provider
+}
+
+func (opt *ReattachOption) configureAdd(conf *addConfig) {
+	conf.reattachInfo = opt.info
+}
+
+// Add represents the terraform add subcommand.
```
radeksimko, line 44:
```suggestion
// Add represents the `terraform add` subcommand (added in 1.1.0).
//
// Note that this function signature and behaviour is subject
// to breaking changes including removal of that function
// until final 1.1.0 Terraform version (with this command) is released.
```

radeksimko (changes requested):
Thank you for the PR @magodo 

I wouldn't say it's necessarily a rule that any `tfinstall`-installable version has to be supported in `tfexec` but as long as it's correctly version-gated then I think it's fine! 😄 

The `add` command may still undergo some changes before getting out of alpha, but it's also equally likely that it will just land as is. With that in mind I also proposed a minor change in the comment, so that folks are more aware of potential BCs. That said they probably will be aware anyway as they have to opt into installing alpha version 🤷🏻‍♂️ 

The PR looks pretty good aside from my in-line comment about the `out` flag and the signature. Once that is addressed I'd be happy to merge it.

tfexec/internal/testutil/tfcache.go
magodo: This makes sense, thank you!

tfexec/add.go
magodo: Indeed, this is totally a mistake. Thank you for pointing this out :+1:

magodo:
@radeksimko Thank you for the quick feedback! I've resolved the comments above, please take another look :)

tfexec/add.go
```diff
@@ -0,0 +1,101 @@
...
+// Note that this function signature and behaviour is subject
+// to breaking changes including removal of that function
+// until final 1.1.0 Terraform version (with this command) is released.
+func (tf *Terraform) Add(ctx context.Context, address string, opts ...AddOption) (string, error) {
+	cmd, err := tf.addCmd(ctx, address, opts...)
+	if err != nil {
+		return "", err
+	}
+
+	var outBuf bytes.Buffer
```
radeksimko, line 56: Nitpick: Given that we're returning `string` and never deal with raw bytes we may as well make this [`strings.Builder`](https://pkg.go.dev/strings#Builder)
```suggestion
	var outBuf strings.Builder
```

This admittedly applies to most current `bytes` usage in other places, but that can be addressed in a separate PR.

radeksimko (commented):
Looking good, I'm happy to merge once that last nitpick is addressed 😃

radeksimko (approved):

PR #217, opened 2021-09-02
Update ErrNoInit text

kmoe:
As of https://github.com/hashicorp/terraform/commit/89b05050ec0d1cae275fb9a3072e85364ba37a22 (https://github.com/hashicorp/terraform/pull/29469), Terraform Core now logs the following error when it detects that `terraform init` was required but not run:

```
Please run "terraform init".
```

In versions 0.13-1.0.5, this text reads instead "Error: could not load plugin".

This PR adds the new text to the regex.

Fixes nightly `main` test failure.


kmoe:
Merging this without review to fix nightly test failure since Radek is away.

radeksimko (commented):
LGTM, thanks!

Issue #222, opened 2021-09-26
Support error wrapping for locked state errors


alec-rabold:
First off thank you for maintaining this project! It's such a great tool for automating TF workflows in code.

**Feature**

One addition that I believe would be helpful is a first-class error when there's an issue acquiring a state lock, as well as the `LockInfo` so that a `force-unlock` is possible if appropriate. The driving use-case for us is when the Terraform process unexpectedly exits early, requiring either manual intervention or a separate `stderr` logger to catch this.

I've created a pull request (https://github.com/hashicorp/terraform-exec/pull/221) and was hoping to get some thoughts from you all.

Thanks!

radeksimko:
#221 was just merged

Issue #121, opened 2021-01-15
Enforce version constraints for command options


kmoe:
For each option, check what Terraform version it was added in, and add a `tf.compatible` check enforcing the minimum version. We might want a better framework for doing this than adding logic in each command, but it might not be worth doing that because option compatibility is per command, not per option, and often influences default flag behaviour (see the changes in https://github.com/hashicorp/terraform-exec/pull/120 for an example).

PR #223, opened 2021-09-28
add command for force-unlock

alec-rabold:
Adds command for `terraform force-unlock -force LOCK_ID [DIR]`. Assumes the `-force` option as it's practically required in this context.

Regarding `forceUnlockDirArgMaxVersion`, the [command docs](https://www.terraform.io/docs/cli/commands/force-unlock.html) appear to indicate that the `[DIR]` positional argument is still valid but it was effectively removed around v0.15.0 in [this commit](https://github.com/hashicorp/terraform/commit/ca23a096d8c48544b9bfc6dbf13c66488f9b6964#diff-fa30e1342d858b292ab8a556b5a3bb7406ec8c744d11dbb267c1bda7cfee3df3L33-R34). Just wanted to clarify where that was coming from since it was failing `v0.15.x`+ test cases.

(Related to https://github.com/hashicorp/terraform-exec/issues/222)

### DIFF
--- a/tfexec/force_unlock.go
+++ b/tfexec/force_unlock.go
@@ -1,0 +1,50 @@
+package tfexec
+
+import (
+	"context"
+	"os/exec"
+)
+
+type forceUnlockConfig struct {
+	dir string
+}
+
+var defaultForceUnlockOptions = forceUnlockConfig{}
+
+type ForceUnlockOption interface {
+	configureForceUnlock(*forceUnlockConfig)
+}
+
+func (opt *DirOption) configureForceUnlock(conf *forceUnlockConfig) {
+	conf.dir = opt.path
+}
+
+// ForceUnlock represents the `terraform force-unlock` command
+func (tf *Terraform) ForceUnlock(ctx context.Context, lockID string, opts ...ForceUnlockOption) error {
+	unlockCmd := tf.forceUnlockCmd(ctx, lockID, opts...)
+
+	if err := tf.runTerraformCmd(ctx, unlockCmd); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (tf *Terraform) forceUnlockCmd(ctx context.Context, lockID string, opts ...ForceUnlockOption) *exec.Cmd {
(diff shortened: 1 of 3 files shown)


radeksimko:
FWIW We have an open PR https://github.com/hashicorp/terraform-exec/pull/100 which discusses the `[DIR]` positional argument, but have not yet decided whether/how to proceed with that PR.

Aside from that - do you mind sharing some details about your use case here?

My understanding is that when state remains locked and Terraform returns that error, it is for that exact reason that "something unexpected happened" and essentially *manual* intervention is recommended. i.e. I would not expect anyone to build any sort of automation script that _automatically_ unlocks the the state.

alec-rabold:
Thanks for the context on `chdir`/`[DIR]`, I think it makes sense to let that PR you mentioned guide the discussion there :+1:

As for use cases, I agree that automatically unlocking the state comes with its fair share of risk and uncertainty. 

Currently we're running Terraform in a Kubernetes ecosystem to support self-service manifests (sort of similar to [aws-controllers-k8s](https://github.com/aws-controllers-k8s)). Every so often we're hit with edge-cases that lock a resource's state like OOM kills or other non-graceful terminations. Since the TF project files aren't meant to be accessed locally, we manually signal that a resource should `force-unlock` by using an annotation on the k8s object (we're running a home-grown `force-unlock` implementation). 

All that being said though, I do think there's valid use-cases where a higher-order construct can decide whether or not to `force-unlock` automatically (e.g. leader-election in our case, possibly).

kmoe (approved):
I'm happy with this. The handling of `[DIR]` is consistent with other commands, and, as you note, the Terraform docs, although its use will produce an error in v0.15+. We could add a `tf.compatible` for it here, or we could wait for it to be caught in https://github.com/hashicorp/terraform-exec/issues/121. I don't think it matters too much.

Await re-review by @radeksimko before merge.

radeksimko (approved):
Yep, I'm happy with this as well 👍🏻 Thanks for sharing the context.

Issue #188, opened 2021-06-07
NewTerraform returns an error without attempting to locate an appropriate binary on the system PATH


suzuki-shunsuke:
NewTerraform is described as the following.

https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.13.3/tfexec#NewTerraform

> If a blank execPath is supplied, NewTerraform will attempt to locate an appropriate binary on the system PATH.

But actually if execPath is empty NewTerraform returns an error without attempting to locate an appropriate binary on the system PATH.

suzuki-shunsuke:
I have sent a pull request to fix this issue. #187
This pull request changes NewTerraform to find `terraform` command by exec.LookPath if execPath is empty.

PR #243, opened 2021-11-10
NewTerraform: Update comment to reflect reality

radeksimko:
This updates the inline comment as pointed out and requested in https://github.com/hashicorp/terraform-exec/pull/187#issuecomment-898073188

Closes https://github.com/hashicorp/terraform-exec/issues/188

cc @suzuki-shunsuke


radeksimko:
As this is just a comment update, I'm going to merge it, but reviews are certainly still welcomed.

PR #245, opened 2021-11-12
Add support for the `workspace show` command

### LINKED ISSUE #239: Support `terraform workspace show`
Add support for `terraform workspace show` command.

lafentres:
## Description

This PR adds support for the `workspace show` command. 

Closes #239 

## Related links

- [workspace show documentation](https://www.terraform.io/docs/cli/commands/workspace/show.html)

### DIFF
--- a/tfexec/workspace_show.go
+++ b/tfexec/workspace_show.go
@@ -1,0 +1,21 @@
+package tfexec
+
+import (
+	"context"
+	"strings"
+)
+
+// WorkspaceShow represents the workspace show subcommand to the Terraform CLI.
+func (tf *Terraform) WorkspaceShow(ctx context.Context) (string, error) {
+	workspaceShowCmd := tf.buildTerraformCmd(ctx, nil, "workspace", "show", "-no-color")
+
+	var outBuffer strings.Builder
+	workspaceShowCmd.Stdout = &outBuffer
+
+	err := tf.runTerraformCmd(ctx, workspaceShowCmd)
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimSpace(outBuffer.String()), nil
+}
--- a/tfexec/workspace_show.go
+++ b/tfexec/workspace_show.go
@@ -2,20 +2,34 @@ package tfexec
 
 import (
 	"context"
+	"fmt"
+	"os/exec"
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


tfexec/workspace_show.go
```diff
@@ -0,0 +1,21 @@
...
+	var outBuffer strings.Builder
+	workspaceShowCmd.Stdout = &outBuffer
+
+	err := tf.runTerraformCmd(ctx, workspaceShowCmd)
+	if err != nil {
+		return "", err
+	}
+
+	return strings.TrimSpace(outBuffer.String()), nil
+}
```
lafentres, line 21: i didn't break this up into `WorkspaceShow` and `workspaceShowCmd` because it didn't seem like that would be very helpful here since this command doesn't really have options i don't think. i'm happy to break it out that way for consistency or if i've missed something though. just let me know!

radeksimko (commented):
This LGTM, my only one reservation is that we aim to also more gracefully handle version differences in tfexec and this command in particular was introduced in `v0.10` https://github.com/hashicorp/terraform/commit/9d7fce2f69123100afd43be27a110f8431fbb43b

Would you therefore mind adding a version check, similar to the one we have for `fmt` here?
https://github.com/hashicorp/terraform-exec/blob/750bc90690f17802eb236adf7336553ea947bcbe/tfexec/fmt.go#L129-L132

FYI There's a more long-term issue tracking this problem for all commands https://github.com/hashicorp/terraform-exec/issues/121

lafentres:
> This LGTM, my only one reservation is that we aim to also more gracefully handle version differences in tfexec and this command in particular was introduced in `v0.10` [hashicorp/terraform@9d7fce2](https://github.com/hashicorp/terraform/commit/9d7fce2f69123100afd43be27a110f8431fbb43b)
> 
> Would you therefore mind adding a version check, similar to the one we have for `fmt` here?
> 
> https://github.com/hashicorp/terraform-exec/blob/750bc90690f17802eb236adf7336553ea947bcbe/tfexec/fmt.go#L129-L132
> 
> FYI There's a more long-term issue tracking this problem for all commands #121

excellent point! i've added a version check for this command and a little bit of extra testing :]

radeksimko (approved):
LGTM :shipit:

PR #251, opened 2021-11-22
Adding Taint/UnTaint command support

rambabuiitk:
Adding output name option support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ FEATURES:
  - Add support for `workspace delete` command [GH-212]
  - Add support for `workspace show` command [GH-245]
  - Add support for `force-unlock` command [GH-223]
+ - Add support for `taint`/`untaint` command [GH-251]
 
 # 0.15.0 (October 05, 2021)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ FEATURES:
  - Add support for `workspace delete` command [GH-212]
  - Add support for `workspace show` command [GH-245]
  - Add support for `force-unlock` command [GH-223]
- - Add support for `taint`/`untaint` command [GH-251]
 
 # 0.15.0 (October 05, 2021)
 
--- a/tfexec/internal/e2etest/untaint_test.go
+++ b/tfexec/internal/e2etest/untaint_test.go
@@ -28,7 +28,7 @@ func TestUntaint(t *testing.T) {
 
 		err = tf.Untaint(context.Background(), "null_resource.foo")
 		if err != nil {
-			t.Fatalf("error running UnTaint: %s", err)
+			t.Fatalf("error running Untaint: %s", err)
 		}
 	})
 }
--- a/tfexec/internal/e2etest/untaint_test.go
+++ b/tfexec/internal/e2etest/untaint_test.go
@@ -9,7 +9,7 @@ import (
 	"github.com/hashicorp/terraform-exec/tfexec"
 )
 
-func TestUnTaint(t *testing.T) {
+func TestUntaint(t *testing.T) {
 	runTest(t, "basic", func(t *testing.T, tfv *version.Version, tf *tfexec.Terraform) {
 		err := tf.Init(context.Background())
 		if err != nil {
@@ -26,7 +26,7 @@ func TestUnTaint(t *testing.T) {
 			t.Fatalf("error running Taint: %s", err)
 		}
 
-		err = tf.UnTaint(context.Background(), "null_resource.foo")
+		err = tf.Untaint(context.Background(), "null_resource.foo")
 		if err != nil {
 			t.Fatalf("error running UnTaint: %s", err)
 		}
--- a/tfexec/output.go
+++ b/tfexec/output.go
@@ -34,7 +34,6 @@ type OutputMeta struct {
 
 // Output represents the terraform output subcommand.
 func (tf *Terraform) Output(ctx context.Context, opts ...OutputOption) (map[string]OutputMeta, error) {
-
 	outputCmd := tf.outputCmd(ctx, opts...)
 
 	outputs := map[string]OutputMeta{}
@@ -52,6 +51,7 @@ func (tf *Terraform) outputCmd(ctx context.Context, opts ...OutputOption) *exec.
 	for _, o := range opts {
 		o.configureOutput(&c)
 	}
+
 	args := []string{"output", "-no-color", "-json"}
 
 	// string opts: only pass if set
--- a/tfexec/output.go
+++ b/tfexec/output.go
@@ -8,6 +8,7 @@ import (
 
 type outputConfig struct {
 	state string
+	name  string
 	json  bool
 }
 
@@ -22,6 +23,10 @@ func (opt *StateOption) configureOutput(conf *outputConfig) {
 	conf.state = opt.path
 }
 
+func (opt *OutputNameOption) configureOutput(conf *outputConfig) {
+	conf.name = opt.name
+}
+
 // OutputMeta represents the JSON output of 'terraform output -json',
 // which resembles state format version 3 due to a historical accident.
 // Please see hashicorp/terraform/command/output.go.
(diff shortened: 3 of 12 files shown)


rambabuiitk:
@radeksimko new PR

rambabuiitk:
@radeksimko any chance I could get the review? the winbuildtest is failing, not sure why though...I couldn't find the reason from the logs...Is that mandatory?

rambabuiitk:
@lafentres Can I get the review for this? I see the issue assigned to you.

tfexec/taint.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (tf *Terraform) taintCmd(ctx context.Context, address string, opts ...TaintOption) *exec.Cmd {
+	c := defaultTaintOptions
+
+	for _, o := range opts {
+		o.configureTaint(&c)
+	}
+
+	args := []string{"taint", "-allow-missing", "-lock-timeout=0s", "-lock=true"}
```
radeksimko, line 36:
```suggestion
	args := []string{"taint", "-no-color"}
```
Given that our intentions behind tfexec is to make it a relatively universal library supporting different use cases, I don't think we should be making any assumptions about the flags here and rely on CLI defaults instead. The only flag which would be basically universally desirable in any automation is `no-color`.

It would be reasonable to introduce each flag as `TaintOption` though, along with `-ignore-remote-version`. Do you mind adding these?

tfexec/untaint.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (tf *Terraform) unTaintCmd(ctx context.Context, address string, opts ...UnTaintOption) *exec.Cmd {
+	c := defaultUnTaintOptions
+
+	for _, o := range opts {
+		o.configureUnTaint(&c)
+	}
+
+	args := []string{"untaint", "-allow-missing", "-lock-timeout=0s", "-lock=true", "-no-color"}
```
radeksimko, line 36:
```suggestion
	args := []string{"untaint", "-no-color"}
```

Similar to above - I don't think we should be setting these flags to any particular value by default and rely on CLI defaults instead.

tfexec/untaint.go
```diff
@@ -0,0 +1,46 @@
...
+type UnTaintOption interface {
+	configureUnTaint(*untaintConfig)
+}
+
+func (opt *StateOption) configureUnTaint(conf *untaintConfig) {
+	conf.state = opt.path
+}
+
+// Untaint represents the terraform untaint subcommand.
+func (tf *Terraform) UnTaint(ctx context.Context, address string, opts ...UnTaintOption) error {
```
radeksimko, line 24:
```suggestion
func (tf *Terraform) Untaint(ctx context.Context, address string, opts ...UnTaintOption) error {
```
Nitpick - I would say that this name would be justified in case the command name was e.g. `un-taint`.
Usually (from my own experience) every 2nd and other capital in `PascalCase` or `camelCase` represents a new word or part of word that is somehow separated from the preceding one.

`untaint` is a whole word though, so I think we can just avoid that extra capital in the middle.

tfexec/output.go
```diff
@@ -34,23 +39,35 @@ type OutputMeta struct {
 
 // Output represents the terraform output subcommand.
 func (tf *Terraform) Output(ctx context.Context, opts ...OutputOption) (map[string]OutputMeta, error) {
-	outputCmd := tf.outputCmd(ctx, opts...)
```
radeksimko, line 37: This looks like a reasonable change, but do you mind raising a separate PR for it or explain why does it really need to be part of this PR? 

It would make it much easier to review and ensure that any side effects of that change get the right attention if it was a separate PR.

radeksimko (changes requested):
Hi @rambabuiitk 
Thanks for your contribution!

Aside from my in-line comments I will say that we aim to also more gracefully handle version differences in tfexec and `taint` in particular was introduced in `v0.4.1` https://github.com/hashicorp/terraform/commit/4ec31ecb9506a794f0c53f90d0cc529bef1aa70d
`untaint` was introduced in `v0.6.13` https://github.com/hashicorp/terraform/commit/c7f5450a964e0fa89a02303232972b4777e6ec63

Would you therefore mind adding a version check, similar to the one we have for `fmt` here?
https://github.com/hashicorp/terraform-exec/blob/750bc90690f17802eb236adf7336553ea947bcbe/tfexec/fmt.go#L129-L132

Having checks for the individual flags would also be nice to have, but I won't block the PR on those.

radeksimko:
As for the failing Windows job I believe that's unrelated to your PR. I noticed we had a similar timeout-triggered failure on `main` recently: https://app.circleci.com/pipelines/github/hashicorp/terraform-exec/1334/workflows/d059b80e-43e4-4ea8-905f-165efaec939f/jobs/16081

I will look into raising the timeout in a separate PR.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ FEATURES:
  - Add support for `workspace delete` command [GH-212]
  - Add support for `workspace show` command [GH-245]
  - Add support for `force-unlock` command [GH-223]
+ - Add support for `taint`/`untaint` command [GH-251]
```
radeksimko, line 17:
```suggestion
```

We usually update the changelog separately from PRs for various reasons - conflicts is one of them.

tfexec/output.go
```diff
@@ -34,6 +34,7 @@ type OutputMeta struct {
 
 // Output represents the terraform output subcommand.
 func (tf *Terraform) Output(ctx context.Context, opts ...OutputOption) (map[string]OutputMeta, error) {
+
```
radeksimko, line 37:
```suggestion
```

tfexec/output.go
```diff
@@ -51,7 +52,6 @@ func (tf *Terraform) outputCmd(ctx context.Context, opts ...OutputOption) *exec.
 	for _, o := range opts {
 		o.configureOutput(&c)
 	}
-
```
radeksimko, line 54: Looks like a stray line deletion 👀

tfexec/untaint.go
```diff
@@ -0,0 +1,51 @@
...
+	"context"
+	"fmt"
+	"os/exec"
+)
+
+type untaintConfig struct {
+	state string
+}
+
+var defaultUnTaintOptions = untaintConfig{}
```
radeksimko, line 13:
```suggestion
var defaultUntaintOptions = untaintConfig{}
```

tfexec/untaint.go
```diff
@@ -0,0 +1,51 @@
...
+
+// OutputOption represents options used in the Output method.
+type UnTaintOption interface {
```
radeksimko, line 16:
```suggestion
type UntaintOption interface {
```

tfexec/untaint.go
```diff
@@ -0,0 +1,51 @@
...
+	configureUnTaint(*untaintConfig)
+}
+
+func (opt *StateOption) configureUnTaint(conf *untaintConfig) {
```
radeksimko, line 20:
```suggestion
func (opt *StateOption) configureUntaint(conf *untaintConfig) {
```

tfexec/untaint.go
```diff
@@ -0,0 +1,51 @@
...
+func (tf *Terraform) Untaint(ctx context.Context, address string, opts ...UnTaintOption) error {
+	err := tf.compatible(ctx, tf0_6_13, nil)
+	if err != nil {
+		return fmt.Errorf("untaint was first introduced in Terraform 0.6.13: %w", err)
+	}
+	unTaintCmd := tf.unTaintCmd(ctx, address, opts...)
+	return tf.runTerraformCmd(ctx, unTaintCmd)
+}
+
+func (tf *Terraform) unTaintCmd(ctx context.Context, address string, opts ...UnTaintOption) *exec.Cmd {
```
radeksimko, line 34:
```suggestion
func (tf *Terraform) untaintCmd(ctx context.Context, address string, opts ...UnTaintOption) *exec.Cmd {
```

tfexec/untaint_test.go
```diff
@@ -0,0 +1,41 @@
+package tfexec
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hashicorp/terraform-exec/tfexec/internal/testutil"
+)
+
+func TestUnTaintCmd(t *testing.T) {
```
radeksimko, line 10:
```suggestion
func TestUntaintCmd(t *testing.T) {
```

tfexec/untaint_test.go
```diff
@@ -0,0 +1,41 @@
...
+	tf, err := NewTerraform(td, tfVersion(t, testutil.Latest013))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// empty env, to avoid environ mismatch in testing
+	tf.SetEnv(map[string]string{})
+
+	t.Run("defaults", func(t *testing.T) {
+		unTaintCmd := tf.unTaintCmd(context.Background(), "aws_instance.foo")
```
radeksimko, line 22:
```suggestion
		untaintCmd := tf.unTaintCmd(context.Background(), "aws_instance.foo")
```

tfexec/untaint_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+		assertCmd(t, []string{
+			"untaint",
+			"-no-color",
+			"aws_instance.foo",
+		}, nil, unTaintCmd)
+	})
+
+	t.Run("override all defaults", func(t *testing.T) {
+		unTaintCmd := tf.unTaintCmd(context.Background(), "aws_instance.foo", State("teststate"))
```
radeksimko, line 32:
```suggestion
		untaintCmd := tf.unTaintCmd(context.Background(), "aws_instance.foo", State("teststate"))
```

radeksimko (changes requested):
Thank you for making these changes promptly.

I left you some more comments in-line.

Also - as suggested - do you mind adding each `taint` and `untaint` flag as an options?

tfexec/internal/e2etest/untaint_test.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+
+		err = tf.Taint(context.Background(), "null_resource.foo")
+		if err != nil {
+			t.Fatalf("error running Taint: %s", err)
+		}
+
+		err = tf.Untaint(context.Background(), "null_resource.foo")
+		if err != nil {
+			t.Fatalf("error running UnTaint: %s", err)
```
radeksimko, line 31:
```suggestion
			t.Fatalf("error running Untaint: %s", err)
```

radeksimko (approved):
LGTM, thanks, I'll just apply that one last minor suggestion and get it merged.

PR #256, opened 2021-12-02
Adding name option for output command.

rambabuiitk:
Updated tests

### DIFF
--- a/tfexec/options.go
+++ b/tfexec/options.go
@@ -214,6 +214,14 @@ func Out(path string) *OutOption {
 	return &OutOption{path}
 }
 
+type OutputNameOption struct {
+	name string
+}
+
+func OutputName(name string) *OutputNameOption {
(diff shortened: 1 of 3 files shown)


rambabuiitk:
@radeksimko As discuss separated the output option to this PR...Wondering if I could get a review for this

radeksimko (changes requested):
Hi @rambabuiitk 
This looks reasonable in principle, but unfortunately as you noticed the JSON output for `output -json NAME` differs significantly from `output -json` in that it doesn't contain any metadata about the type or sensitivity.

Therefore I'm afraid this is unsuitable for integration into the existing `Output()` interface.

The JSON output of `terraform output -json` is also now considered protected under Terraform v1 guarantees, so unlikely to change until at least v2.

--- 

Do you mind describing the use case which isn't served by the existing `Output()` / `output -json` and map key access?

PR #259, opened 2021-12-10
Added workspace support

nmiculinic:


sergiorivera50:
Is this dead?

fsaintjacques:
@radeksimko this is a blocker for us. We would like to run tfexec *concurrently* for different workspaces. Depending on the state of `.terraform/environment` makes this impossible without a lock. This PR would allow us to do it concurrently for independent workspaces.

PR #268, opened 2022-01-18
fix force-copy

### LINKED ISSUE #262: Init() always gives `-force-copy` as an argument
Init() always gives `-force-copy` flag into the `terraform init` command.

https://github.com/hashicorp/terraform-exec/blob/v0.15.0/tfexec/init.go#L119

In my understanding, this means that the terraform always (re)initialization the backend  with `-migrate-state` even if `-reconfigure` is given.

https://www.terraform.io/cli/commands/init#backend-initialization

I consider that Init() should  remove `-force-copy` if it gives `-reconfigure` flag to the terraform command.

chanwit:
This PR fixes unconfigurable force-copy, and change its default value to be false.

Fixes #262

### DIFF
--- a/tfexec/init.go
+++ b/tfexec/init.go
@@ -52,6 +52,10 @@ func (opt *DirOption) configureInit(conf *initConfig) {
 	conf.dir = opt.path
 }
 
+func (opt *ForceCopyOption) configureInit(conf *initConfig) {
+	conf.forceCopy = opt.forceCopy
+}
+
 func (opt *FromModuleOption) configureInit(conf *initConfig) {
 	conf.fromModule = opt.source
 }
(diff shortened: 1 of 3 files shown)


Nalum:
Any chance of getting this in?

kmoe (approved):
Thanks for this PR - this was indeed an oversight and I agree with the fix.

The test failures should be fixed if you rebase your branch on `main`. I'll merge once tests pass.

chanwit:
Thank you @kmoe 
I'm working on it.

chanwit:
Hi again @kmoe 
It seems all tests passed, except the mac OS as it could not spin up the test environment.

kmoe:
The `macosbuildtest` failure is unrelated to this PR and fixed in `main`, so I am merging this.

PR #270, opened 2022-01-20
Run force-unlock with -no-color flag

HadrienPatte:
This adds the `-no-color` flag by default to the `force-unlock` command.

Without this patch we can get `ForceUnlock` error:
```
exit status 1
[31mFailed to unlock state: Lock ID should be numerical value, got 'ID'[0m[0m
```
With this patch the error becomes:
```
exit status 1
Failed to unlock state: Lock ID should be numerical value, got 'ID'
```
(I purposefully ran `ForceUnlock` with an invalid argument here to have it error out)


kmoe (approved):
LGTM. I'm going to follow up with some tests, and will merge this PR when they are ready.

PR #271, opened 2022-01-23
Add support for setting plugin cache directory

alec-rabold:
- Adds support for `TF_PLUGIN_CACHE_DIR` environment variable (introduced in [0.10.7](https://github.com/hashicorp/terraform/blob/v0.11/CHANGELOG.md#0107-october-2-2017))

I think this could also be an option on the `init` command -- whichever y'all think would be best!

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -15,17 +15,18 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


sbocinec:
I wonder, other than the failing tests and making it go 1.18 compatible, is there any other blocker to add support for this ENV variable? It would be very helpful

PR #275, opened 2022-01-29
Add JSON option for "plan" and "apply"

### LINKED ISSUE #183: Plan/Apply `-json` support missing
Plans and applies now support streaming json with the `-json` flag. Adding support for the flag may not be enough--including streamed data structures from `stdout` would be preferred.

kylecarbs:
Fixes #183.


johnstcn:
69231a8  adds `terraform` stdout to tfexec output.

PR #284, opened 2022-03-16
Bump Go module compatibility to Go 1.17

bflad:
This will match the README documentation, especially now that Go 1.18 is released.

Updated via:

```
go mod edit -go=1.17
go mod tidy
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -57,6 +57,24 @@ jobs:
       - go_e2e_test
 
   # previous stable Go version
+  go116_build:
+    docker:
+      - image: circleci/golang:1.16
+    steps:
+      - checkout
+      - go_build
+  go116_test:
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
Thanks for the PR!

Since the motivating factor was release of Go 1.18, is it worth also to start testing against 1.18? 
https://github.com/hashicorp/terraform-exec/commit/a430f848bca79f6a201fb0c37534b985580c5b60

Not a blocker for the PR though.

bflad:
@radeksimko pushed up the CircleCI configuration updates, please take a look!

bflad:
Ah, will need to revert that as its not available yet:

```
Error response from daemon: manifest for circleci/golang:1.18 not found: manifest unknown: manifest unknown
```

PR #286, opened 2022-03-25
tests: remove testify dependency

kmoe:
This was only used in one place, and we don't need it.

Somewhat baffled as to why `go mod tidy` doesn't remove the `testify` lines from `go.sum`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,6 @@ require (
 	github.com/hashicorp/hc-install v0.3.1
 	github.com/hashicorp/terraform-json v0.13.0
 	github.com/sergi/go-diff v1.2.0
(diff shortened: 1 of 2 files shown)


radeksimko (approved):

bflad:
@kmoe here's why 😉 

```
❯ go mod why -m github.com/stretchr/testify
# github.com/stretchr/testify
github.com/hashicorp/terraform-exec/tfexec/internal/e2etest
github.com/hashicorp/terraform-exec/tfexec/internal/e2etest.test
github.com/sergi/go-diff/diffmatchpatch
github.com/sergi/go-diff/diffmatchpatch.test
github.com/stretchr/testify/assert
```

PR #289, opened 2022-04-13
deps: Remove github.com/andybalholm/crlf

bflad:
Replaced via trivial `strings` handling as the strings are not large.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,6 @@ module github.com/hashicorp/terraform-exec
 go 1.17
 
 require (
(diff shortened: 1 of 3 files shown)


radeksimko (approved):
I'm confused as to why the [upstream implementation](https://github.com/andybalholm/crlf/blob/master/transform.go) looks so complicated/verbose. I can only guess it's for performance? Either way - we only use it in tests anyway - so as long as Windows tests pass this LGTM :shipit:

PR #291, opened 2022-04-13
tfexec: Add (Terraform).SetLog() method

### LINKED ISSUE #290: Support Configurable TF_LOG environment variable >= Terraform CLI 0.15
## Description

Post Terraform CLI 0.15, setting log levels outside TRACE is now considered reliable. Currently, terraform-exec hardcodes the equivalent of `TF_LOG=TRACE` when using [`(tfexec.Terraform).SetLogPath()`](https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.16.0/tfexec#Terraform.SetLogPath):

https://github.com/hashicorp/terraform-exec/blob/1bbcd0b684b56bc77004691687cbeeae25c67555/tfexec/cmd.go#L145-L154

Consumers, such as the terraform-plugin-sdk/v2 acceptance testing framework should be able to pass along a provider developer's desire for DEBUG or higher logs only from Terraform CLI.

## Proposal

- Add `tfexec.Terraform` type `log string` field.
- Add `(tfexec.Terraform).SetLog(string) error` method, which sets the `log` field if Terraform CLI >= 0.15.
- Update `(tfexec.Terraform).SetLogPath(string)` to set the `log` field to `TRACE`, unless it is already set (this will ensure `SetLog()` or `SetLogPath()` can be called in any order and preserve the previous behavior)
- Update `(tfexec.Terraform).buildEnv()` logic to call `env[logEnvVar] = tf.log` instead of `env[logEnvVar] = "TRACE"`

## References

- Related: https://github.com/hashicorp/terraform-exec/issues/101

bflad:
Closes #290

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -149,8 +149,7 @@ func (tf *Terraform) buildEnv(mergeEnv map[string]string) []string {
 		env[logPathEnvVar] = ""
 	} else {
 		env[logPathEnvVar] = tf.logPath
-		// Log levels other than TRACE are currently unreliable, the CLI recommends using TRACE only.
-		env[logEnvVar] = "TRACE"
+		env[logEnvVar] = tf.log
 	}
 
 	// constant automation override env vars
--- a/tfexec/terraform.go
+++ b/tfexec/terraform.go
@@ -48,9 +48,14 @@ type Terraform struct {
 	skipProviderVerify bool
 	env                map[string]string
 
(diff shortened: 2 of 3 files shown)


bflad:
Rebased and pushed up environment variable testing for both `SetLog` and `SetLogPath`. Please let me know if you'd like anything else. 👍

radeksimko (approved):
LGTM, thanks!

PR #292, opened 2022-04-15
apply: allow use of -destroy flag for compatible terraform versions

rstandt:
Implement the ApplyOption interface for the DestroyFlagOption
struct which enables the user to run `terraform apply -destroy`
as they would using the terraform binary directly by calling
`func (tf *Terraform) Apply`.

### DIFF
--- a/tfexec/internal/e2etest/apply_test.go
+++ b/tfexec/internal/e2etest/apply_test.go
@@ -15,6 +15,10 @@ import (
 	"github.com/hashicorp/terraform-exec/tfexec/internal/testutil"
 )
 
+var (
+	applyDestroyMinVersion = version.Must(version.NewVersion("0.15.2"))
+)
+
 func TestApply(t *testing.T) {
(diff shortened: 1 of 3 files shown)


rstandt:
Resolves https://github.com/hashicorp/terraform-exec/issues/293

tfexec/internal/e2etest/apply_test.go
```diff
@@ -9,6 +9,10 @@ import (
 	"github.com/hashicorp/terraform-exec/tfexec"
 )
 
+var (
+	validateMinApplyDestroyVersion = version.Must(version.NewVersion("0.15.2"))
```
kmoe, line 13:
```suggestion
	applyDestroyMinVersion = version.Must(version.NewVersion("0.15.2"))
```

kmoe (approved):
Fab, thank you. Just one change so the min version var matches our naming convention.

tfexec/internal/e2etest/apply_test.go
pwhitehead00: would be great to merge this, i can fork the author's branch if that's helpful

PR #301, opened 2022-05-09
Canceling Terraform Command

lornasong:
This PR:
- Fixes an issue where canceling context was not killing the Terraform command as expected
- Adds a workaround for a secondary Golang issue where canceling a command when Stdout and Stderr are set can cause hanging.
    - Note: I was a little surprised that that the hanging issue doesn't impact all versions of TF i.e. some versions of TF pass but others fail

Commits (see message for more details, sample logs, and progression of working through issue)
1. Add timeout `TestContext_sleepTimeoutExpired` to check if the Terraform Apply is killed within a reasonable timeframe of the cancel (test fails because terraform apply is not killed)
2. Add fix so that Terraform Apply is killed. This introduces the hanging issue (test fails because of hanging)
3. Add workaround for hanging (test passes)

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -1,9 +1,11 @@
 package tfexec
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


lornasong:
Closed since it is a sub-set of changes in https://github.com/hashicorp/terraform-exec/pull/299

PR #304, opened 2022-05-17
make tests pass (or skip) on darwin/arm64

radeksimko:
**Why?**
Prior to this patch, many tests would fail for the same reason - attempt to install older version of Terraform which isn't available for darwin/arm64.

For example:
```
--- FAIL: TestDestroyCmd (0.08s)
    terraform_test.go:164: caching exec "v:0.12.31" in dir "/var/folders/9w/pddq2x656j76vxvsbhcs4nwr0000gq/T/tfinstall1718612104/v-0.12.31"
panic: error installing terraform "v:0.12.31": no ZIP archive found for terraform <nil> darwin/arm64 [recovered]
	panic: error installing terraform "v:0.12.31": no ZIP archive found for terraform <nil> darwin/arm64

goroutine 99 [running]:
testing.tRunner.func1.2({0x102b2d280, 0x1400038cb80})
	/opt/homebrew/Cellar/go/1.18.1/libexec/src/testing/testing.go:1389 +0x1c8
testing.tRunner.func1()
	/opt/homebrew/Cellar/go/1.18.1/libexec/src/testing/testing.go:1392 +0x384
panic({0x102b2d280, 0x1400038cb80})
	/opt/homebrew/Cellar/go/1.18.1/libexec/src/runtime/panic.go:838 +0x204
github.com/hashicorp/terraform-exec/tfexec/internal/testutil.(*TFCache).find(0x140002ca0c0, 0x14000003d40, {0x140004cc690, 0x9}, 0x14000077ea0)
	/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-exec/tfexec/internal/testutil/tfcache_find.go:47 +0x374
github.com/hashicorp/terraform-exec/tfexec/internal/testutil.(*TFCache).Version(0x102b2d280?, 0x14000003d40, {0x1029fefa5, 0x7})
	/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-exec/tfexec/internal/testutil/tfcache.go:61 +0x90
github.com/hashicorp/terraform-exec/tfexec.tfVersion(0x14000003d40?, {0x1029fefa5?, 0x14000067f48?})
	/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-exec/tfexec/terraform_test.go:164 +0x88
github.com/hashicorp/terraform-exec/tfexec.TestDestroyCmd(0x14000003d40)
	/Users/radeksimko/gopath/src/github.com/hashicorp/terraform-exec/tfexec/destroy_test.go:13 +0x44
testing.tRunner(0x14000003d40, 0x102bf6820)
	/opt/homebrew/Cellar/go/1.18.1/libexec/src/testing/testing.go:1439 +0x110
created by testing.(*T).Run
	/opt/homebrew/Cellar/go/1.18.1/libexec/src/testing/testing.go:1486 +0x300
FAIL	github.com/hashicorp/terraform-exec/tfexec	6.097s
```

**Alternatives**
 - We could build Terraform as part of running relevant tests. That would likely add to the test time, increase the complexity of already complex matrix of versions throughout all packages and risk that a particular version may simply not work on darwin/arm64 as it was never properly tested.
 - We could update _all_ tests to use v1.0.2+ which _have_ darwin/arm64 builds, but there are still tests for older TF versions which may be useful to run.

**Solution**

I tried to strike the balance and bump TF version where it was possible/easy to do and skipped test where it wasn't.

--- 

I wish we could run any arm64 tests in CI, but AFAIK GitHub Actions don't offer that (yet).

### DIFF
--- a/tfexec/init_test.go
+++ b/tfexec/init_test.go
@@ -73,3 +73,57 @@ func TestInitCmd_v012(t *testing.T) {
 		}, nil, initCmd)
 	})
 }
+
+func TestInitCmd_v1(t *testing.T) {
+	td := t.TempDir()
+
+	tf, err := NewTerraform(td, tfVersion(t, testutil.Latest_v1))
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// empty env, to avoid environ mismatch in testing
+	tf.SetEnv(map[string]string{})
+
+	t.Run("defaults", func(t *testing.T) {
+		// defaults
+		initCmd, err := tf.initCmd(context.Background())
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		assertCmd(t, []string{
+			"init",
+			"-no-color",
+			"-force-copy",
+			"-input=false",
+			"-backend=true",
+			"-get=true",
+			"-upgrade=false",
+		}, nil, initCmd)
+	})
+
+	t.Run("override all defaults", func(t *testing.T) {
+		initCmd, err := tf.initCmd(context.Background(), Backend(false), BackendConfig("confpath1"), BackendConfig("confpath2"), FromModule("testsource"), Get(false), PluginDir("testdir1"), PluginDir("testdir2"), Reconfigure(true), Upgrade(true), Dir("initdir"))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		assertCmd(t, []string{
+			"init",
+			"-no-color",
+			"-force-copy",
+			"-input=false",
+			"-from-module=testsource",
+			"-backend=false",
+			"-get=false",
+			"-upgrade=true",
+			"-reconfigure",
+			"-backend-config=confpath1",
+			"-backend-config=confpath2",
+			"-plugin-dir=testdir1",
+			"-plugin-dir=testdir2",
+			"initdir",
+		}, nil, initCmd)
+	})
+}
--- a/tfexec/init_test.go
+++ b/tfexec/init_test.go
@@ -2,12 +2,17 @@ package tfexec
 
 import (
 	"context"
+	"runtime"
 	"testing"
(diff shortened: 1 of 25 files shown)


tfexec/init_test.go
```diff
@@ -2,12 +2,17 @@ package tfexec
...
 import (
 	"context"
+	"runtime"
 	"testing"
 
 	"github.com/hashicorp/terraform-exec/tfexec/internal/testutil"
 )
 
-func TestInitCmd(t *testing.T) {
+func TestInitCmd_v012(t *testing.T) {
```
bflad, line 11: Nit: Should this also have a v1 to ensure coverage? Seems like it'd be nice if the tests each had a mini-matrix of Terraform versions and expected behaviors, although it'd be a lot of effort with (potentially) little gain.

tfexec/terraform_test.go
```diff
@@ -273,6 +273,10 @@ func TestSetLogPath(t *testing.T) {
 }
 
 func TestCheckpointDisablePropagation(t *testing.T) {
```
bflad, line 275: Nit: Should there be a v1 test to ensure coverage?

tfexec/workspace_show_test.go
```diff
@@ -3,10 +3,15 @@ package tfexec
 import (
 	"context"
 	"errors"
+	"runtime"
 	"testing"
 )
 
 func TestWorkspaceShowCmd(t *testing.T) {
```
bflad, line 10: Nit: Similarly regarding coverage

bflad (approved):
Looks good to me, just some small non-blocking considerations 🚀

kmoe (approved):

PR #305, opened 2022-05-18
Enable e2e testing against custom Terraform executable

kmoe:
If the env var `TFEXEC_E2ETEST_TERRAFORM_PATH` is set to the path of a valid local Terraform executable, when e2e tests are run, only tests appropriate to that executable's version will be run.

This involves both the `runTest()` and `runTestVersions()` codepaths, since both are used in e2e tests.

The code here is quite awkward due to the need to support multiple codepaths. Happy to refactor more if we prefer.

### DIFF
--- a/tfexec/internal/e2etest/util_test.go
+++ b/tfexec/internal/e2etest/util_test.go
@@ -45,6 +45,11 @@ func runTest(t *testing.T, fixtureName string, cb func(t *testing.T, tfVersion *
 	// valid Terraform executable, only tests appropriate to that
 	// executable's version will be run.
 	if localBinPath := os.Getenv("TFEXEC_E2ETEST_TERRAFORM_PATH"); localBinPath != "" {
+		// By convention, every new Terraform struct is given a clean
+		// temp dir, even if we are only invoking tf.Version(). This
+		// prevents any possible confusion that could result from
+		// reusing an os.TempDir() (for example) that already contained
+		// Terraform files.
 		td, err := ioutil.TempDir("", "tf")
 		if err != nil {
 			t.Fatalf("error creating temporary test directory: %s", err)
--- a/tfexec/internal/e2etest/util_test.go
+++ b/tfexec/internal/e2etest/util_test.go
@@ -41,6 +41,32 @@ func runTest(t *testing.T, fixtureName string, cb func(t *testing.T, tfVersion *
 		versions = strings.Split(override, ",")
 	}
 
+	// If the env var TFEXEC_E2ETEST_TERRAFORM_PATH is set to the path of a
+	// valid Terraform executable, only tests appropriate to that
+	// executable's version will be run.
+	if localBinPath := os.Getenv("TFEXEC_E2ETEST_TERRAFORM_PATH"); localBinPath != "" {
+		td, err := ioutil.TempDir("", "tf")
+		if err != nil {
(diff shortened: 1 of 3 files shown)


tfexec/internal/e2etest/util_test.go
```diff
@@ -41,6 +41,32 @@ func runTest(t *testing.T, fixtureName string, cb func(t *testing.T, tfVersion *
...
+	// valid Terraform executable, only tests appropriate to that
+	// executable's version will be run.
+	if localBinPath := os.Getenv("TFEXEC_E2ETEST_TERRAFORM_PATH"); localBinPath != "" {
+		td, err := ioutil.TempDir("", "tf")
+		if err != nil {
+			t.Fatalf("error creating temporary test directory: %s", err)
+		}
+		t.Cleanup(func() {
+			os.RemoveAll(td)
+		})
```
radeksimko, line 54: Is there a reason we have to _create_ a directory for the execution of `terraform version`? Wouldn't it suffice to use  [`os.TempDir()`](https://pkg.go.dev/os#TempDir)?

tfexec/internal/e2etest/util_test.go
```diff
@@ -63,12 +89,12 @@ func runTestVersions(t *testing.T, versions []string, fixtureName string, cb fun
 				os.RemoveAll(td)
 			})
 
-			// TODO: do this in a cleaner way than string comparison?
 			var execPath string
-			switch {
-			case strings.HasPrefix(tfv, "refs/"):
+			if localBinPath := os.Getenv("TFEXEC_E2ETEST_TERRAFORM_PATH"); localBinPath != "" {
+				execPath = localBinPath
+			} else if strings.HasPrefix(tfv, "refs/") {
```
radeksimko, line 95: I'm wondering if long-term we could just get rid of this entirely and assume that if Terraform is built from scratch, then it's done outside of tfexec and path to the binary is passed via `TFEXEC_E2ETEST_TERRAFORM_PATH`? 🤔 

Not sure how many tests actually still use this though.

radeksimko (approved):
Just two minor questions, but not necessarily blocking.

tfexec/internal/e2etest/util_test.go
kmoe: Perhaps this is a little paranoid. The failure case for using `os.TempDir()` is that it contains some Terraform configuration files that somehow in a future version of Terraform affect the result of `terraform version`. I agree this is pretty unlikely. I do, however, like our convention that during tests, every time a new `Terraform{}` is initialised it is in a totally clean directory.

radeksimko: Good point - I _think_ it _already_ behaves differently depending on path, because it shows versions of installed providers. So - good call to execute in a clean directory.

Maybe just leave a comment to explain _why_ we're doing that for future ourselves?

tfexec/internal/e2etest/util_test.go
radeksimko: Yeah - I admit it would be a lot to ask as part of this PR, but it just seemed appealing from the perspective of reducing the surface of the library, if building it outside is an option now with your new ENV variable.

kmoe: Opened https://github.com/hashicorp/terraform-exec/issues/318

PR #307, opened 2022-05-20
ci: Test with -race flag

radeksimko:
This ensures that we don't reintroduce race conditions, which were recently addressed in https://github.com/hashicorp/terraform-exec/pull/299

`go test -race` tends to take longer than without the flag, so I also pre-emptively increased the timeout for the E2E test run.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -26,12 +26,12 @@ commands:
   go_unit_test:
     steps:
       - run:
-          command: go test $(go list ./... | grep -v /tfexec/internal/e2etest)
+          command: go test -race $(go list ./... | grep -v /tfexec/internal/e2etest)
   go_e2e_test:
     steps:
       - run:
-          no_output_timeout: 20m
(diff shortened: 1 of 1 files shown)


kmoe (approved):
Sounds good. 30 minutes of tests...!

radeksimko:
> 30 minutes of tests...!

I think in reality they'd more likely finish much faster, especially with the default `count=1`, but I reached ~15min mark with `count=5` locally.

Looking at the CI now:

 - Windows: 8mins vs 11mins
 - macOS: 8mins vs 9mins
 - Linux: 3mins vs 8mins

I'm happy to dial it down, but I also think that unless they take _significantly_ longer to finish I'd much rather interpret this as a data point, than a hard/loud failure. I mean - the most likely correction we'd apply should this happen is bumping the timeout.

If performance is what we're trying to measure here, then benchmarking seems like a much better tool than our existing tests.

PR #316, opened 2022-06-16
Fix for state_push: path for statefile is reqired for run state push cmd

aav66:
_terrafrom state push command_ expect statefile path.
If it not specified it will fail with "Exactly one argument expected"

documentation link: https://www.terraform.io/cli/commands/state/push#usage


aav66:
go117_test_main-1 failed with `panic: error installing terraform "gitref:refs/heads/main": unable to clone terraform from "https://github.com/hashicorp/terraform.git": context deadline exceeded`
Just need to be restarted.

kmoe (approved):
Thanks for doing this. Even basic e2e tests would have caught this bug - my mistake in not ensuring they were added initially.

PR #320, opened 2022-06-24
Add basic e2e tests for state push/pull

kmoe:
Extremely basic e2e tests for state push/pull. These would have caught the bug fixed in https://github.com/hashicorp/terraform-exec/pull/316.

I wonder if `StatePull` should also return a state struct instead of a `string`.


radeksimko (approved):
Thanks for the test.

> I wonder if StatePull should also return a state struct instead of a string.

Maybe? Depends on how we represent state elsewhere. Also there's a potential concern with scale if the state is big - maybe `io.Writer` interface which leaves the writing to the consumer is better?

Similarly `StatePush` could also take `io.Reader` and pass the content as STDIN.

I think we'll need to rethink these interfaces, along with `Format*` methods for v1.

kmoe:
>Maybe? Depends on how we represent state elsewhere.

`Show()` returns a `*tfjson.State`, which I think is reasonable. We should at least be consistent, I think.

>Also there's a potential concern with scale if the state is big - maybe io.Writer interface which leaves the writing to the consumer is better?

Yes, good point... Opened https://github.com/hashicorp/terraform-exec/issues/321

>I think we'll need to rethink these interfaces, along with Format* methods for v1.
Do you mean possibly removing `FormatString()`? We currently have both this and `Format()` which uses `io.Reader`/`Writer`.

PR #329, opened 2022-08-18
json log for plan and apply

aav66:
Hello!
Would like to merge support for json logs for plan and apply.
@kmoe Could you review it? If anything else needed please let me know.

It is similar for #275 but with out changes to stdout and stderr writers.


aav66:
@kmoe Is any change to merge this feature? Is any thing else needed to done?

PR #332, opened 2022-08-30
Support graceful cancellation

simon0191:
This PR adds support for graceful cancellation and a `GracefulShutdownTimeout` option for apply and destroy commands.

Currently when the context is cancelled a SIGKILL is sent to Terraform. This is the default behaviour of `exec.CommandContext` . The SIGKILL forcefully kills the Terraform process and therefore it doesn’t start a graceful shutdown which sometimes ends up with the state lock not being released.

The proposed fix is that on context cancellation, a SIGINT is sent so that Terraform starts a graceful shutdown with the hope of releasing the lock and allowing providers to gracefully terminate, and then, after `GracefulShutdownTimeout` if the command hasn’t finished, then a SIGKILL is sent to forcefully terminate the process.

---

When a command is started, the command that is created is run without a context (equivalent to passing `context.Background()`), and in a concurrent go routine the cancellation of the parent context is monitored.

When the parent context is cancelled, a SIGINT is sent to the command. After `GracefulShutdownTimeout` passes, if the command hasn't finished, a SIGKILL is sent to the command to forcefully kill it.

The `GracefulShutdownTimeout` is also useful so that the caller of tf-exec is able to configure for how long it's willing to wait for the cancellation, which depends on how each provider handles cancellations.

## Refs

- https://github.com/hashicorp/terraform-plugin-sdk/issues/141
- https://github.com/hashicorp/terraform-provider-aws/issues/4177
- https://github.com/hashicorp/terraform-provider-aws/issues/15090

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ func (tf *Terraform) Refresh(ctx context.Context, opts ...RefreshCmdOption) erro
 
 func (tf *Terraform) refreshCmd(ctx context.Context, opts ...RefreshCmdOption) (*exec.Cmd, error) {
 	...
-  	return tf.buildTerraformCmd(ctx, mergeEnv, args...), nil
+  	return tf.buildTerraformCmd( mergeEnv, args...), nil
 }
 ```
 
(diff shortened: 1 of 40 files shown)


simon0191:
Closing in favor of https://github.com/hashicorp/terraform-exec/pull/334 which is a much simpler and nicer implementation

PR #334, opened 2022-08-31
tfexec: add InterruptChannel option to allow a user triggered interrupt

twmb:
This is an alternative to #332, adding an option that enables the user
to implement graceful shutdowns for Apply and Destroy operations.

Rather than adding an option that changes the behavior of the input
context, we instead add an option that specifically sends an interrupt
to the terraform process. The input context behavior remains unchanged.

This requires the caller to do a bit more orchestration work for
timeouts, but keeps context truer to the "abandon work" intent. This
also allows users to force quit _even if_ they are in the middle of a
graceful shutdown, rathern than having one behavior mutually exclusive
with the other.

# Refs

* https://github.com/hashicorp/terraform-plugin-sdk/issues/141
* https://github.com/hashicorp/terraform-provider-aws/issues/4177
* https://github.com/hashicorp/terraform-provider-aws/issues/15090


tfexec/cmd_default.go
```diff
@@ -47,6 +48,19 @@ func (tf *Terraform) runTerraformCmd(ctx context.Context, cmd *exec.Cmd) error {
 		return tf.wrapExitError(ctx, err, "")
 	}
 
+	if interruptCh := ctx.Value(interruptContext); interruptCh != nil {
+		exited := make(chan struct{})
+		defer close(exited)
+		go func() {
+			select {
+			case <-interruptCh.(<-chan struct{}):
+				cmd.Process.Signal(os.Interrupt)
```
simon0191, line 57: I think SIGINT is not available in Windows and that's what's making the test to fail 🤔

simon0191: I'm not sure why cmd_linux can't be used for macos as well?

simon0191: I've made cmd_linux -> cmd_unix as the implementation should work in all unix systems
https://github.com/hashicorp/terraform-exec/pull/334/commits/44715fe64ecdc5a92e4bf7916c99951df0df3293

bilalcaliskan:
is there any update about this feature request and this PR? cheers

esacteksab:
is there any update about this feature request and this PR? cheers

radeksimko:
Hello @twmb 
Apologies for the radio silence here.

We decided to take a slightly different approach in https://github.com/hashicorp/terraform-exec/pull/512 to solve the problem which hopefully further simplifies the UX and leverages the native API of `os/exec`. Admittedly this API wasn't available at the time you raised this PR (https://go-review.googlesource.com/c/go/+/401835 was submitted in October 2022).

Either way, feel free to upgrade to [v0.23.0](https://github.com/hashicorp/terraform-exec/releases/tag/v0.23.0) which was just released and contains the linked change.

PR #345, opened 2022-11-16
Fix resource leak

nicolai86:
We ran into issues with `*terraform.Apply` where individual executions are interrupted and retried.

If an execution is interrupted without sending `SIGINT` terraform losses changes to the backend, breaking subsequent retries.

Some context when this happens:
- we use the eks module [eks](registry.terraform.io/terraform-aws-modules/eks/aws)
- we use the pg backend for terraform

This change fixes the issue for us.

I'll try to add a minimal reproducable setup to show case in a bit.

### DIFF
--- a/tfexec/cmd_linux.go
+++ b/tfexec/cmd_linux.go
@@ -9,6 +9,9 @@ import (
 )
 
 func (tf *Terraform) runTerraformCmd(ctx context.Context, cmd *exec.Cmd) error {
+	ctx, done := context.WithCancel(ctx)
+	defer done()
+
 	var errBuf strings.Builder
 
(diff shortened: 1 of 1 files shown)


radeksimko:
Hello @nicolai86
Thanks for the report and the contribution.

I believe https://github.com/hashicorp/terraform-exec/pull/512 which leverages more of the native `os/exec` API, should address the mentioned issues, including the one from your TODO. The default timeout after cancellation (which now sends SIGINT instead of SIGKILL) is 60s after which SIGKILL is sent. This is also adjustable as per [v0.23.0 Changelog](https://github.com/hashicorp/terraform-exec/releases/tag/v0.23.0).

Feel free to upgrade to the most recent version and let us know if this is still an issue.

PR #350, opened 2023-01-09
ci: (re)enable releasing via GitHub Action workflow

radeksimko:
This follows a guide in https://github.com/hashicorp/signore/blob/main/docs/SIGNING_GIT_COMMITS.md and https://github.com/hashicorp/signore/blob/main/docs/SIGNING_IN_CI.md

## TODO

 - [x] register new bot GitHub account for holding new GPG key
 - [x] update script to use new email
 - [x] get a new signore signer
 - [x] upload new pubkey to the bot GitHub account
 - [x] ensure the correct `SIGNORE_SIGNER` is set
 - [x] ensure the correct `SIGNORE_CLIENT` & `SIGNORE_SECRET` are set
 - [x] Add release instructions to CONTRIBUTING.md

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,34 @@
+name: Release
+
+permissions:
+  contents: write # for modifying version in-place & pushing tags
+  packages: read # for downloading signore docker image
+
+on:
+  workflow_dispatch:
+
+env:
+  GOPROXY: https://proxy.golang.org/
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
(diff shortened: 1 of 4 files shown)


dbanck (approved):
LGTM 👍 

I assume using `git commit --gpg-sign="${SIGNORE_SIGNER}"`, removes the need to update the git config like outlined in the signore docs:
```
git config --global user.signingkey $your_signore_signer
git config --global commit.gpgsign true
```

radeksimko:
> I assume using git commit --gpg-sign="${SIGNORE_SIGNER}", removes the need to update the git config like outlined in the signore docs:

Yes, that is my understanding as well. The `git config` options would mean "always sign every commit/tag", whereas we're just being more explicit there by passing the flags to `git commit` and `git tag`.

Issue #353, opened 2023-01-10
Add functions that run Terraform commands with `-json` flag 


bendbennett:
### Use Case

There are a number of instances in which it would be useful to execute Terraform commands (e.g., _terraform apply_) with the `-json` flag. For instance, in order to be able to robustly [identify warnings](https://github.com/hashicorp/terraform-plugin-testing/issues/16) that have been emitted during the execution of a Terraform command, it would be useful to be able to capture the JSON output from the command and subsequently parse and interrogate it. 

### Proposal

Add the following functions to _terraform exec_:

- `ApplyJSON(ctx context.Context, w io.Writer, opts ...ApplyOption) error`
- `PlanJSON(ctx context.Context, w io.Writer, opts ...PlanOption) (bool, error)`
- `RefreshJSON(ctx context.Context, w io.Writer, opts ...RefreshCmdOption) error`
- `DestroyJSON(ctx context.Context, w io.Writer, opts ...DestroyOption) error`

Note that ` Validate(ctx context.Context) (*tfjson.ValidateOutput, error)` is already running _terraform refresh_  with the `-json` flag

bflad:
Related: https://github.com/hashicorp/terraform-exec/issues/183

PR #354, opened 2023-01-10
Adding ApplyJSON(), DestroyJSON(), PlanJSON() and RefreshJSON() functions

bendbennett:
Closes: #353

### DIFF
--- a/tfexec/apply.go
+++ b/tfexec/apply.go
@@ -120,18 +120,6 @@ func (tf *Terraform) ApplyJSON(ctx context.Context, w io.Writer, opts ...ApplyOp
 	return tf.runTerraformCmd(ctx, cmd)
 }
 
-func (tf *Terraform) applyJSONCmd(ctx context.Context, opts ...ApplyOption) (*exec.Cmd, error) {
-	cmd, err := tf.applyCmd(ctx, opts...)
-	if err != nil {
-		return nil, err
-	}
-
-	cmd.Args = append(cmd.Args[:3], cmd.Args[2:]...)
-	cmd.Args[2] = "-json"
-
-	return cmd, nil
-}
-
 func (tf *Terraform) applyCmd(ctx context.Context, opts ...ApplyOption) (*exec.Cmd, error) {
 	c := defaultApplyOptions
 
@@ -139,6 +127,32 @@ func (tf *Terraform) applyCmd(ctx context.Context, opts ...ApplyOption) (*exec.C
 		o.configureApply(&c)
 	}
 
+	args, err := tf.buildApplyArgs(ctx, c)
+	if err != nil {
+		return nil, err
+	}
+
+	return tf.buildApplyCmd(ctx, c, args)
+}
+
+func (tf *Terraform) applyJSONCmd(ctx context.Context, opts ...ApplyOption) (*exec.Cmd, error) {
+	c := defaultApplyOptions
+
+	for _, o := range opts {
+		o.configureApply(&c)
+	}
+
+	args, err := tf.buildApplyArgs(ctx, c)
+	if err != nil {
+		return nil, err
+	}
+
+	args = append(args, "-json")
+
+	return tf.buildApplyCmd(ctx, c, args)
+}
+
+func (tf *Terraform) buildApplyArgs(ctx context.Context, c applyConfig) ([]string, error) {
 	args := []string{"apply", "-no-color", "-auto-approve", "-input=false"}
 
 	// string opts: only pass if set
@@ -184,6 +198,10 @@ func (tf *Terraform) applyCmd(ctx context.Context, opts ...ApplyOption) (*exec.C
 		}
 	}
 
+	return args, nil
+}
+
+func (tf *Terraform) buildApplyCmd(ctx context.Context, c applyConfig, args []string) (*exec.Cmd, error) {
 	// string argument: pass if set
 	if c.dirOrPlan != "" {
 		args = append(args, c.dirOrPlan)
--- a/tfexec/apply.go
+++ b/tfexec/apply.go
@@ -101,7 +101,15 @@ func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
 }
 
 // ApplyJSON represents the Terraform apply subcommand with the `-json` flag.
+// Using the `-json` flag will result in
+// [machine-readable](https://developer.hashicorp.com/terraform/internals/machine-readable-ui)
+// JSON being written to the supplied `io.Writer`.
 func (tf *Terraform) ApplyJSON(ctx context.Context, w io.Writer, opts ...ApplyOption) error {
+	err := tf.compatible(ctx, tf0_15_3, nil)
+	if err != nil {
+		return fmt.Errorf("terraform apply -json was added in 0.15.3: %w", err)
+	}
+
 	tf.SetStdout(w)
 
 	cmd, err := tf.applyJSONCmd(ctx, opts...)
--- a/tfexec/apply.go
+++ b/tfexec/apply.go
@@ -3,6 +3,7 @@ package tfexec
 import (
 	"context"
 	"fmt"
+	"io"
 	"os/exec"
 	"strconv"
 )
@@ -90,13 +91,37 @@ func (opt *ReattachOption) configureApply(conf *applyConfig) {
 	conf.reattachInfo = opt.info
 }
 
-// Apply represents the terraform apply subcommand.
+// Apply represents the Terraform apply subcommand.
 func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
 	cmd, err := tf.applyCmd(ctx, opts...)
 	if err != nil {
 		return err
 	}
 	return tf.runTerraformCmd(ctx, cmd)
+}
+
+// ApplyJSON represents the Terraform apply subcommand with the `-json` flag.
+func (tf *Terraform) ApplyJSON(ctx context.Context, w io.Writer, opts ...ApplyOption) error {
+	tf.SetStdout(w)
+
+	cmd, err := tf.applyJSONCmd(ctx, opts...)
+	if err != nil {
+		return err
+	}
+
+	return tf.runTerraformCmd(ctx, cmd)
+}
+
+func (tf *Terraform) applyJSONCmd(ctx context.Context, opts ...ApplyOption) (*exec.Cmd, error) {
+	cmd, err := tf.applyCmd(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	cmd.Args = append(cmd.Args[:3], cmd.Args[2:]...)
+	cmd.Args[2] = "-json"
+
+	return cmd, nil
 }
 
 func (tf *Terraform) applyCmd(ctx context.Context, opts ...ApplyOption) (*exec.Cmd, error) {
--- a/tfexec/apply.go
+++ b/tfexec/apply.go
@@ -103,7 +103,8 @@ func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
 // ApplyJSON represents the terraform apply subcommand with the `-json` flag.
 // Using the `-json` flag will result in
 // [machine-readable](https://developer.hashicorp.com/terraform/internals/machine-readable-ui)
-// JSON being written to the supplied `io.Writer`.
+// JSON being written to the supplied `io.Writer`. ApplyJSON is likely to be
+// removed in a future major version in favour of Apply returning JSON by default.
 func (tf *Terraform) ApplyJSON(ctx context.Context, w io.Writer, opts ...ApplyOption) error {
 	err := tf.compatible(ctx, tf0_15_3, nil)
 	if err != nil {
--- a/tfexec/apply.go
+++ b/tfexec/apply.go
@@ -91,7 +91,7 @@ func (opt *ReattachOption) configureApply(conf *applyConfig) {
 	conf.reattachInfo = opt.info
 }
 
-// Apply represents the Terraform apply subcommand.
+// Apply represents the terraform apply subcommand.
 func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
 	cmd, err := tf.applyCmd(ctx, opts...)
 	if err != nil {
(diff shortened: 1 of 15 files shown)


tfexec/apply.go
```diff
@@ -99,6 +100,30 @@ func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
 	return tf.runTerraformCmd(ctx, cmd)
 }
 
+// ApplyJSON represents the Terraform apply subcommand with the `-json` flag.
```
bflad, line 103: These new commands could likely use a link to https://developer.hashicorp.com/terraform/internals/machine-readable-ui in the Go documentation (or something to help consumers try to use it).

tfexec/apply_test.go
```diff
@@ -65,3 +65,63 @@ func TestApplyCmd(t *testing.T) {
 		}, nil, applyCmd)
 	})
 }
+
+func TestApplyJSONCmd(t *testing.T) {
```
bflad, line 69: There's also the integration testing folder if you want to add end-to-end testing code. 👍

tfexec/apply.go
```diff
@@ -99,6 +100,30 @@ func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
...
+func (tf *Terraform) ApplyJSON(ctx context.Context, w io.Writer, opts ...ApplyOption) error {
+	tf.SetStdout(w)
```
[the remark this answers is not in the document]
bendbennett: Have added e2e test coverage.

tfexec/destroy.go
```diff
@@ -95,6 +96,30 @@ func (tf *Terraform) Destroy(ctx context.Context, opts ...DestroyOption) error {
 	return tf.runTerraformCmd(ctx, cmd)
 }
 
+// DestroyJSON represents the Terraform destroy subcommand with the `-json` flag.
+func (tf *Terraform) DestroyJSON(ctx context.Context, w io.Writer, opts ...DestroyOption) error {
+	tf.SetStdout(w)
```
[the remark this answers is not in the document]
bendbennett: Have added e2e test coverage.

tfexec/plan.go
```diff
@@ -108,6 +109,42 @@ func (tf *Terraform) Plan(ctx context.Context, opts ...PlanOption) (bool, error)
...
+// PlanJSON executes `terraform plan` with the specified options as well as the
+// `-json` flag and waits for it to complete.
+//
+// The returned boolean is false when the plan diff is empty (no changes) and
+// true when the plan diff is non-empty (changes present).
+//
+// The returned error is nil if `terraform plan` has been executed and exits
+// with either 0 or 2.
+func (tf *Terraform) PlanJSON(ctx context.Context, w io.Writer, opts ...PlanOption) (bool, error) {
+	tf.SetStdout(w)
```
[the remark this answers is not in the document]
bendbennett: Have added e2e test coverage.

tfexec/refresh.go
```diff
@@ -78,6 +79,30 @@ func (tf *Terraform) Refresh(ctx context.Context, opts ...RefreshCmdOption) erro
 	return tf.runTerraformCmd(ctx, cmd)
 }
 
+// RefreshJSON represents the Terraform refresh subcommand with the `-json` flag.
+func (tf *Terraform) RefreshJSON(ctx context.Context, w io.Writer, opts ...RefreshCmdOption) error {
+	tf.SetStdout(w)
```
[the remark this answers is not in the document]
bendbennett: Have added e2e test coverage.

tfexec/apply.go
bendbennett: Have linked to https://developer.hashicorp.com/terraform/internals/machine-readable-ui  in the Go docs.

tfexec/apply_test.go
bendbennett: Have added e2e tests.

tfexec/internal/e2etest/util_test.go
```diff
@@ -76,6 +76,43 @@ func runTest(t *testing.T, fixtureName string, cb func(t *testing.T, tfVersion *
 	runTestVersions(t, versions, fixtureName, cb)
 }
 
+func runTestWithVersions(t *testing.T, fixtureName string, versions []string, cb func(t *testing.T, tfVersion *version.Version, tf *tfexec.Terraform)) {
```
bflad, line 79: Should `runTest()` be consolidated to just this?

```go
func runTest(t *testing.T, fixtureName string, cb func(t *testing.T, tfVersion *version.Version, tf *tfexec.Terraform)) {
	t.Helper()

	versions := []string{
		testutil.Latest011,
		testutil.Latest012,
		testutil.Latest013,
		testutil.Latest014,
		testutil.Latest015,
		testutil.Latest_v1,
		testutil.Latest_v1_1,
	}

	if override := os.Getenv("TFEXEC_E2ETEST_VERSIONS"); override != "" {
		versions = strings.Split(override, ",")
	}

	runTestWithVersions(t, fixtureName, versions, cb)
}
```

tfexec/apply.go
```diff
@@ -99,6 +100,38 @@ func (tf *Terraform) Apply(ctx context.Context, opts ...ApplyOption) error {
...
+}
+
+func (tf *Terraform) applyJSONCmd(ctx context.Context, opts ...ApplyOption) (*exec.Cmd, error) {
+	cmd, err := tf.applyCmd(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	cmd.Args = append(cmd.Args[:3], cmd.Args[2:]...)
+	cmd.Args[2] = "-json"
```
radeksimko, line 130: I'm assuming the slice contains 2 arguments for sure `[]string{"terraform", "apply", ...}` and while it's _likely_ there will be other arguments after `apply`, I'm not sure it's a good idea to rely on it as a change in the underlying `applyCmd()` could cause panic here.

TL;DR Is there a strong reason for replacing the argument, instead of just appending `-json` to the end?

tfexec/destroy.go
```diff
@@ -95,6 +96,38 @@ func (tf *Terraform) Destroy(ctx context.Context, opts ...DestroyOption) error {
...
+}
+
+func (tf *Terraform) destroyJSONCmd(ctx context.Context, opts ...DestroyOption) (*exec.Cmd, error) {
+	cmd, err := tf.destroyCmd(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	cmd.Args = append(cmd.Args[:3], cmd.Args[2:]...)
+	cmd.Args[2] = "-json"
```
radeksimko, line 126: I'm assuming the slice contains 2 arguments for sure `[]string{"terraform", "destroy", ...}` and while it's _likely_ there will be other arguments after `destroy`, I'm not sure it's a good idea to rely on it as a change in the underlying `destroyCmd()` could cause panic here.

TL;DR Is there a strong reason for replacing the argument, instead of just appending `-json` to the end?

tfexec/internal/e2etest/plan_test.go
```diff
@@ -45,5 +48,44 @@ func TestPlanWithState(t *testing.T) {
...
+func TestPlanJSON_TF014AndEarlier(t *testing.T) {
+	versions := []string{testutil.Latest011, testutil.Latest012, testutil.Latest013, testutil.Latest014}
+
+	runTestWithVersions(t, "basic", versions, func(t *testing.T, tfv *version.Version, tf *tfexec.Terraform) {
+		err := tf.Init(context.Background())
+		if err != nil {
+			t.Fatalf("error running Init in test directory: %s", err)
+		}
 
+		re := regexp.MustCompile("terraform apply -json was added in 0.15.3")
```
radeksimko, line 62:
```suggestion
		re := regexp.MustCompile("terraform plan -json was added in 0.15.3")
```

tfexec/internal/e2etest/refresh_test.go
```diff
@@ -27,3 +30,37 @@ func TestRefresh(t *testing.T) {
...
+func TestRefreshJSON_TF014AndEarlier(t *testing.T) {
+	versions := []string{testutil.Latest011, testutil.Latest012, testutil.Latest013, testutil.Latest014}
+
+	runTestWithVersions(t, "basic", versions, func(t *testing.T, tfv *version.Version, tf *tfexec.Terraform) {
+		err := tf.Init(context.Background())
+		if err != nil {
+			t.Fatalf("error running Init in test directory: %s", err)
+		}
+
+		re := regexp.MustCompile("terraform apply -json was added in 0.15.3")
```
radeksimko, line 43:
```suggestion
		re := regexp.MustCompile("terraform refresh -json was added in 0.15.3")
```

tfexec/refresh.go
```diff
@@ -78,6 +80,38 @@ func (tf *Terraform) Refresh(ctx context.Context, opts ...RefreshCmdOption) erro
...
+}
+
+func (tf *Terraform) refreshJSONCmd(ctx context.Context, opts ...RefreshCmdOption) (*exec.Cmd, error) {
+	cmd, err := tf.refreshCmd(ctx, opts...)
+	if err != nil {
+		return nil, err
+	}
+
+	cmd.Args = append(cmd.Args[:3], cmd.Args[2:]...)
+	cmd.Args[2] = "-json"
```
radeksimko, line 110: I'm assuming the slice contains 2 arguments for sure `[]string{"terraform", "refresh", ...}` and while it's _likely_ there will be other arguments after `refresh`, I'm not sure it's a good idea to rely on it as a change in the underlying `refreshCmd()` could cause panic here.

TL;DR Is there a strong reason for replacing the argument, instead of just appending `-json` to the end?

radeksimko (commented):
Aside from my mostly minor in-line comments I have three higher-level questions:

 - Can you confirm the assumption that with JSON output, the error matching does _not_ work as expected (i.e. we produce generic error types, rather than the ones from https://github.com/hashicorp/terraform-exec/blob/main/tfexec/errors.go )? It's fine if it doesn't - that is one of the reasons we wanted to make it a separate method, but it would be great to have that theory confirmed + more importantly have it documented - ideally both in the comment above the method and in the Changelog. We should have an answer for those who'd rightly be confused about the similar methods.
 - The other reason (in addition to ^) we wanted to make separate `*JSON` methods is that all methods should work with JSON by default. I'm assuming the (in)stability is already communicated through the `0.x.x` version, but is it worth mentioning the (likely) future of these methods in the comment? i.e. that they're most likely to be removed in the future major version in favour of Apply/Refresh/Destroy?
 - Have you tried using this API without buffering? I'm guessing it should work just fine with the help of [`io.Pipe`](https://pkg.go.dev/io#Pipe)? Just trying to get an answer for @thrashr888 in https://github.com/hashicorp/terraform-exec/issues/183 who I believe might want to avoid buffering.

Thank you for adding all the tests. I'm assuming the 3 failures are related to the upstream `hc-install` bug, which is something I'm working on, so you can ignore these for now.

tfexec/apply.go
bendbennett: Refactored to split the building of the args into flags that preceed the optional dir/plan which must come after all flags.

tfexec/destroy.go
bendbennett: Refactored to split the building of the args into flags that preceed the optional dir/plan which must come after all flags.

tfexec/refresh.go
bendbennett: Refactored to split the building of the args into flags that preceed the optional dir/plan which must come after all flags.

bendbennett:
> Can you confirm the assumption that with JSON output, the error matching does not work as expected (i.e. we produce generic error types, rather than the ones from https://github.com/hashicorp/terraform-exec/blob/main/tfexec/errors.go )? It's fine if it doesn't - that is one of the reasons we wanted to make it a separate method, but it would be great to have that theory confirmed + more importantly have it documented - ideally both in the comment above the method and in the Changelog. We should have an answer for those who'd rightly be confused about the similar methods.

With the `...JSON()` functions the original errors (e.g., `ErrVersionMismatch`) are preserved. 
We make use of this in _terraform-plugin-testing_ to fallback to calling the original functions (e.g., `Apply()`) if `ApplyJSON()` generates a `ErrVersionMismatch` error. 

We're using code such as the following in _terraform-plugin-testing_ to detect such errors:
```go
	if err != nil {
		target := &tfexec.ErrVersionMismatch{}
		if errors.As(err, &target) {
			...
```

image

See https://github.com/hashicorp/terraform-plugin-testing/pull/17/files#diff-3894fabafd3f4b1861dd167c6194c6d79e5824d2bec30d37b257d80a824255c2R35 for an example implementation.

@radeksimko given the original errors are preserved can you confirm that the documentation and changelog _do not_ need altering?


> The other reason (in addition to ^) we wanted to make separate *JSON methods is that all methods should work with JSON by default. I'm assuming the (in)stability is already communicated through the 0.x.x version, but is it worth mentioning the (likely) future of these methods in the comment? i.e. that they're most likely to be removed in the future major version in favour of Apply/Refresh/Destroy?

I have updated the Go docs for `ApplyJSON()`, `DestroyJSON()`, `PlanJSON()` and `RefreshJSON()` to reflect that the `...JSON()` functions will likely be removed in future in favour of updating `Apply()` etc to return JSON.


> Have you tried using this API without buffering? I'm guessing it should work just fine with the help of [io.Pipe](https://pkg.go.dev/io#Pipe)? Just trying to get an answer for @thrashr888 in https://github.com/hashicorp/terraform-exec/issues/183 who I believe might want to avoid buffering.

I have tested the API without buffering (i.e., using [io.Pipe](https://pkg.go.dev/io#Pipe)). 

The following test produces the output below:
```go
func TestApplyJSONCmdPipe(t *testing.T) {
	td := t.TempDir()

	tf, err := NewTerraform(td, tfVersion(t, testutil.Latest_v1))
	if err != nil {
		t.Fatal(err)
	}

	// empty env, to avoid environ mismatch in testing
	tf.SetEnv(map[string]string{})

	pipeReader, pipeWriter := io.Pipe()
	defer pipeWriter.Close()

	go func() {
		defer pipeReader.Close()
		if _, err := io.Copy(os.Stdout, pipeReader); err != nil {
			log.Fatal(err)
		}
	}()

	err = tf.ApplyJSON(context.Background(), pipeWriter)
	if err == nil {
		t.Fatal("error expected, got nil")
	}
}
```

```
{"@level":"info","@message":"Terraform 1.0.11","@module":"terraform.ui","@timestamp":"2023-01-23T09:38:37.659360Z","terraform":"1.0.11","type":"version","ui":"0.1.0"}
{"@level":"error","@message":"Error: No configuration files","@module":"terraform.ui","@timestamp":"2023-01-23T09:38:37.660401Z","diagnostic":{"severity":"error","summary":"No configuration files","detail":"Apply requires configuration to be present. Applying without a configuration would mark everything for destruction, which is normally not what is desired. If you would like to destroy everything, run 'terraform destroy' instead."},"type":"diagnostic"}
--- PASS: TestApplyJSONCmdPipe (9.76s)
```

PR #358, opened 2023-02-07
Add support for `metadata functions -json` command

dbanck:
This PR adds support for the new `terraform metadata functions -json` command, which was introduced in https://github.com/hashicorp/terraform/pull/32487. We need this to obtain the function signatures within terraform-ls.

Depends on:
* https://github.com/hashicorp/terraform-json/pull/68 

## TODO

- [x] Bump terraform-json version after it was released
- [x] Add an E2E test? ~~That might require the Terraform 1.4 release~~ It's working for `main` https://github.com/hashicorp/terraform-exec/actions/runs/4176231939/jobs/7232445810#step:4:910

### DIFF
--- a/tfexec/metadata_functions.go
+++ b/tfexec/metadata_functions.go
@@ -1,0 +1,39 @@
+package tfexec
+
+import (
+	"context"
+	"fmt"
+	"os/exec"
+
+	tfjson "github.com/hashicorp/terraform-json"
+)
+
+// MetadataFunctions represents the terraform metadata functions -json subcommand.
+func (tf *Terraform) MetadataFunctions(ctx context.Context) (*tfjson.MetadataFunctions, error) {
+	err := tf.compatible(ctx, tf1_4_0, nil)
+	if err != nil {
+		return nil, fmt.Errorf("terraform metadata functions was added in 1.4.0: %w", err)
+	}
+
+	functionsCmd := tf.metadataFunctionsCmd(ctx)
+
+	var ret tfjson.MetadataFunctions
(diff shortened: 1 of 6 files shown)


tfexec/metadata_functions.go
```diff
@@ -0,0 +1,39 @@
...
+	var ret tfjson.MetadataFunctions
+	err = tf.runTerraformCmdJSON(ctx, functionsCmd, &ret)
+	if err != nil {
+		return nil, err
+	}
+
+	err = ret.Validate()
+	if err != nil {
+		return nil, err
+	}
```
radeksimko, line 29:
```suggestion
```
I believe the validation already runs as part of unmarshaling, so we'd run it twice in this case, wouldn't we?

https://github.com/hashicorp/terraform-json/blob/6ebf04334b66853bb051c5300861aa970f0660e8/metadata.go#L68

I'm assuming https://github.com/hashicorp/terraform-exec/blob/615df2ea4708c84d1facc097d6e2fa158c01ade8/tfexec/providers_schema.go#L20-L23 is only there because earlier versions tfjson didn't do that and we then forgot to remove it.

radeksimko (approved):
Confirming that the output is coming through in the E2E test: https://github.com/hashicorp/terraform-exec/actions/runs/4176231939/jobs/7232437983#step:4:913

👍🏻 LGTM :shipit:

tfexec/metadata_functions.go
dbanck: You're right. We're running validation twice.

Similar story for the provider schemas (my source of inspiration) and show command.

PR #359, opened 2023-02-08
fix issues with version-specific tests when run against a local binary

radditude:
When running e2e tests against a locally built Terraform binary set with `TFEXEC_E2ETEST_TERRAFORM_PATH`, the `runTestWithVersions` helper (introduced in #354) has some surprising behavior: it throws away the list of versions that was passed in and runs the test against the local binary instead. This seems fairly undesirable and was likely not an intended effect of that prior PR, since the tests using that helper are testing version-specific behavior (and thus are prone to fail when a local bin path is set!).

This PR proposes moving that version override behavior into `runTest`, where it was previously, and combining `runTestWithVersions` and `runTestVersions` into one test helper that runs the given test function against _only_ the versions passed in.

Closes https://github.com/hashicorp/terraform/issues/32639

**Potential future enhancement:**

If it would be valuable to have the ability to run some tests against a specified list of versions _as well as_ the local binary if it exists, probably a clearer way to do that would be to add something like a `runTestWithVersionsAndLatest` helper that reads from `TFEXEC_E2ETEST_TERRAFORM_PATH` and appends the local binary's version instead of overwriting the others.

### DIFF
--- a/tfexec/internal/e2etest/apply_test.go
+++ b/tfexec/internal/e2etest/apply_test.go
@@ -29,7 +29,7 @@ func TestApply(t *testing.T) {
 func TestApplyJSON_TF014AndEarlier(t *testing.T) {
 	versions := []string{testutil.Latest011, testutil.Latest012, testutil.Latest013, testutil.Latest014}
 
-	runTestWithVersions(t, "basic", versions, func(t *testing.T, tfv *version.Version, tf *tfexec.Terraform) {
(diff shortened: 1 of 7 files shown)


kmoe (approved):
Thanks @radditude, this is much clearer!

I'm going to merge this now as I think the e2e failures against Terraform `main` and/or 1.4 are resolved by https://github.com/hashicorp/terraform-exec/pull/361.

PR #362, opened 2023-02-09
runTestWithVersions() in CONTRIBUTING.md

kmoe:
The `runTestVersions()` function has been effectively renamed to `runTestWithVersions()` and its behaviour modified by #354 ([e2etest/util_test.go]( https://github.com/hashicorp/terraform-exec/pull/354/files#diff-ea7385346bdbc5b135c051039bb1440eabf601dc7ff447a9a0d119a2f74d97df)) and #359 ([e2etest/util_test.go](https://github.com/hashicorp/terraform-exec/pull/359/files#diff-ea7385346bdbc5b135c051039bb1440eabf601dc7ff447a9a0d119a2f74d97df)) since the contributing docs were written.

Is CONTRIBUTING.md still correct?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,7 +43,7 @@ We aim for full test coverage of all Terraform CLI commands implemented in `tfex
 The following environment variables can be set during testing:
 
  - `TFEXEC_E2ETEST_VERSIONS`: When set to a comma-separated list of version strings, this overrides the default list of Terraform versions for end-to-end tests, and runs those tests against only the versions supplied.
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
I think that a lot of the confusion in tests generally comes from all the abstraction layers which lead to us hard-coding versions we test against, instead of managing versions entirely outside of tests and _skipping_ any tests which we know aren't relevant to run for the given version.

TL;DR I think we should just drop `runTestWithVersions` entirely - but of course that won't be a single LOC change 😅 

Having it correctly documented in the meantime makes sense!

bendbennett (approved):
LGTM 👍

radditude (approved):
Ah, I missed these in #359! 🙌🏻 

To @radeksimko's point, I briefly went down a rabbit hole of trying to write a helper that would allow specifying a version constraint instead of a list of versions, so someone could express something like "run this test against minor versions greater than 1.0" or "run this test on all minor versions between 0.12 and 0.15" without listing out the versions manually. Maybe it would be an interesting path to explore in the future?

PR #364, opened 2023-02-20
scripts/release: update env var for GHA

kmoe:
Note that GHA's `GITHUB_REF` is the whole refspec, not just the branch name, which should still work according to the man page.

### DIFF
--- a/scripts/release/release.sh
+++ b/scripts/release/release.sh
@@ -72,7 +72,7 @@ function commitChanges {
       git tag -a -m "v${TARGET_VERSION}" -s "v${TARGET_VERSION}"
   fi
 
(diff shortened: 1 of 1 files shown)


radeksimko (approved):
Unsure if the "fully qualified" ref is what we want or what we used before

https://docs.github.com/en/actions/learn-github-actions/variables
`GITHUB_REF=refs/heads/<branch_name>`

It may work, but what we might want is `GITHUB_REF_NAME`?

kmoe:
Thanks, I believe `GITHUB_REF` will work (`man git-push`) but `GITHUB_REF_NAME` is what we had before - changing.

PR #365, opened 2023-02-20
ci: Enable token-based auth in release workflow

radeksimko:


radeksimko:
Closing in favour of https://github.com/hashicorp/terraform-exec/pull/366

PR #369, opened 2023-02-20
ci: swap flag order in git commit

radeksimko:
The assumption here is that `SIGNORE_SIGNER` in the script contains a newline, or some other character like `#` near the end which makes it ignore the `-m` flag. Just to rule this theory out (or confirm), I'm proposing different order of these flags in hope to avoid empty commit messages:

https://github.com/hashicorp/terraform-exec/commit/fe344c871e114c54034ca9cb83f83ca23f518278


radeksimko:
This turns out to be an upstream signore bug, patch is now pending review.

PR #372, opened 2023-03-01
bugfix: correctly populate stderr

kmoe:
https://github.com/hashicorp/terraform-exec/pull/352 introduced a bug in which stderr from Terraform commands was not included in the error returned from `tf.Apply()`, etc.

This PR fixes this. Note the inclusion of `errBuf` where `tf.wrapExitError()` was used before.

Also add a test step that fails before this change.

### DIFF
--- a/tfexec/cmd_default.go
+++ b/tfexec/cmd_default.go
@@ -5,6 +5,7 @@ package tfexec
 
 import (
 	"context"
+	"fmt"
 	"os/exec"
 	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


sergiught (approved):
Really appreciate the super quick turnaround @kmoe <3

bendbennett (approved):
LGTM 👍

kmoe:
The registry is serving us a selection of network errors. Rerunning tests.

PR #388, opened 2023-05-02
remove workspace env var and not set it to an empty string.

bjhaid:
Setting the environment variable to an empty string confuses the terraform binary as seen below:

```
TF_WORKSPACE= terraform init -no-color -input=false -get=false -upgrade=false -backend=false

Error: Terraform Cloud returned an unexpected error

invalid value for workspace
```

With this patch the `TF_WORKSPACE` will not exist and avoids the above problem. I have tested this locally with terraform 1.4.5 and it works.

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -169,7 +169,7 @@ func (tf *Terraform) buildEnv(mergeEnv map[string]string) []string {
 	env[automationEnvVar] = "1"
 
 	// force usage of workspace methods for switching
-	env[workspaceEnvVar] = ""
+	delete(env, workspaceEnvVar)
 
 	if tf.disablePluginTLS {
 		env[disablePluginTLSEnvVar] = "1"
--- a/tfexec/cmd_test.go
+++ b/tfexec/cmd_test.go
@@ -46,7 +46,6 @@ func defaultEnv() []string {
 		"TF_LOG_CORE=",
 		"TF_LOG_PATH=",
 		"TF_LOG_PROVIDER=",
-		"TF_WORKSPACE=",
 	}
 }
 


mrak:
Some additional context: the current behavior (explicitly setting `TF_WORKSPACE=""`) breaks tag-selected workspaces using Terraform Cloud such as this:

```hcl
terraform {
  cloud {
    organization = "MyOrg"

    workspaces {
      tags = [
        "owner:my-team",
        "project:my-project",
      ]
    }
  }
}

```

Normally you'd use `terraform workspace select` to select the workspace, but `terraform workspace` cannot be used until `terraform init` has been run. In order for `terraform init` to run you need to either specify `TF_WORKSPACE=my-workspace` or you need to populate `.terraform/environment`.

The bug here is if `TF_WORKSPACE` is defined _regardless of value_ `terraform init` skips any checking of `.terraform/environment`, included if the value of the env var is `""`.

By removing the `TF_WORKSPACE` env var instead of setting it to an empty string allows the automation pattern of using `.terraform/environment` to function as intended.

See the official HashiCorp help article [Selecting a workspace when running Terraform in automation](https://support.hashicorp.com/hc/en-us/articles/360043550953-Selecting-a-workspace-when-running-Terraform-in-automation)

kmoe (approved):
This may well be a bug in the Terraform automation plugin - have let the team know. In any case this is the intended behaviour for terraform-exec. Thanks!

bjhaid:
@kmoe I fixed the tests can you help re-trigger the builds, thanks!

bjhaid:
@kmoe looks like all checks have passed, how do we get this merged and released?

mrak:
Any chance to get a release out with this change? Thanks!

PR #400, opened 2023-07-25
tfexec: Initial test command support

### LINKED ISSUE #398: Consider supporting `terraform test`
Added in Terraform v1.6.0.

Could we test interrupts? https://github.com/hashicorp/terraform/pull/33477

bflad:
Closes #398
Reference: https://github.com/hashicorp/terraform/pull/33454

Adds support for the `terraform test` command, which currently supports JSON machine-readable output and one flag for configuring the tests directory away from the command default of `tests`. The command will return a non-zero status if any of the tests fail, which returns an error back to callers of the `Test` function. If consumers need access to the pass/fail test results, the terraform-json Go module will need to be enhanced to support the test summary JSON, e.g.

```
{"@level":"info","@message":"Failure! 0 passed, 1 failed.","@module":"terraform.ui","@timestamp":"2023-07-25T10:03:42.980799-04:00","test_summary":{"status":"fail","passed":0,"failed":1,"errored":0,"skipped":0},"type":"test_summary"}
```

Output of new end-to-end testing:

```
$ TFEXEC_E2ETEST_VERSIONS=1.5.3,1.6.0-alpha20230719 go test -count=1 -run='TestTest' -v ./tfexec/internal/e2etest
...
--- PASS: TestTest (9.50s)
    --- SKIP: TestTest/test_command_passing-1.5.3 (4.06s)
    --- PASS: TestTest/test_command_passing-1.6.0-alpha20230719 (5.44s)
...
--- PASS: TestTestError (0.48s)
    --- SKIP: TestTestError/test_command_failing-1.5.3 (0.27s)
    --- PASS: TestTestError/test_command_failing-1.6.0-alpha20230719 (0.21s)
```

### DIFF
--- a/tfexec/internal/e2etest/test_test.go
+++ b/tfexec/internal/e2etest/test_test.go
@@ -1,0 +1,56 @@
+// Copyright (c) HashiCorp, Inc.
+// SPDX-License-Identifier: MPL-2.0
+
+package e2etest
+
+import (
+	"context"
+	"io"
+	"regexp"
+	"testing"
+
+	"github.com/hashicorp/go-version"
+
+	"github.com/hashicorp/terraform-exec/tfexec"
+)
+
+var (
+	testMinVersion = version.Must(version.NewVersion("1.6.0"))
+)
+
+func TestTest(t *testing.T) {
+	runTest(t, "test_command_passing", func(t *testing.T, tfv *version.Version, tf *tfexec.Terraform) {
+		// Use Core() to enable pre-release support
+		if tfv.Core().LessThan(testMinVersion) {
+			t.Skip("test command is not available in this Terraform version")
+		}
+
+		err := tf.Test(context.Background(), nil)
+
+		if err != nil {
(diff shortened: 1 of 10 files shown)


kmoe:
Lovely! @liamcervante will be adding terraform-json support for the return types, so those will definitely be usable here.

liamcervante:
> @liamcervante will be adding terraform-json support for the return types, so those will definitely be usable here.

So, terraform-json doesn't actually support the streamed JSON logs for any of the other commands. I'm not sure if I should just go and add them for the testing command only. I'm gathering other opinions on this, is thing something that would be required for this PR? If yes, that'll help me make my decision!

I'm only hesitant as I'm worried about why the streamed logs for the other commands aren't available.

liamcervante:
We've agreed to be reactive here, so happy to merge this as-is. We can come and add more fine-grained info to the outputs as and when we have a concrete use-case.

kmoe:
Lovely, thank you bflad

PR #403, opened 2023-08-24
Escape characters for workspace selection

aanogueira:
This PR aims to fix workspace selection for cases where the namespace starts with special characters, such as `-`.


aanogueira:
Could I please get some review?

ferrastas:
Just a quick example:

```
 $ terraform workspace list
* -dev
  -org
  -stag
```
...
```
$ terraform workspace select -org
Usage: terraform [global options] workspace select NAME

  Select a different Terraform workspace.

Options:

    -or-create=false    Create the Terraform workspace if it doesn't exist.
Error parsing command-line flags: flag provided but not defined: -org
```
...
```
$ terraform workspace select -- -org
Switched to workspace "-org".
```

bflad:
One of the challenges here is that human-readable outputs in Terraform are not protected by compatibility promises, so relying on the existing output might break in future versions.

There are some recent discussions/issues around this:

- https://github.com/hashicorp/terraform/pull/34750#issuecomment-1971779099
- https://github.com/hashicorp/terraform/issues/34775

I think ideally, Terraform CLI would have an enhancement on the existing command, such as a `-json` flag, for machine-readable output. That output (and this library) would be protected from potentially unexpected changes.

PR #418, opened 2023-10-20
Allow using a new process group for the terraform cli child process

kristina-solovyova:
Support using `syscall.SysProcAttr` with `Setpgid: true` on the Cmd struct, which creates a new process group for the child process.   
This effectively separates the child process from the parent process's signal group, so it won't receive the interrupt signal when you interrupt the parent process.  
Motivation - graceful shutdown implementation.  When receiving  CTRL+C the child process running terraform is interrupted, which is undesirable for my use case.

See: https://stackoverflow.com/questions/33165530/prevent-ctrlc-from-interrupting-exec-command-in-golang/33171307#33171307

### DIFF
--- a/tfexec/cmd.go
+++ b/tfexec/cmd.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"syscall"
 
 	"github.com/hashicorp/terraform-exec/internal/version"
 )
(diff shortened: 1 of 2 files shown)


spirit1431007:
How about Windows, we can use [CREATE_NEW_PROCESS_GROUP](https://pkg.go.dev/syscall#CREATE_NEW_PROCESS_GROUP)

radeksimko:
Hello @kristina-solovyova 
Thanks for the contribution.

I believe https://github.com/hashicorp/terraform-exec/pull/512 which leverages more of the native `os/exec` API, should address the mentioned motivation

> Motivation - graceful shutdown implementation. When receiving CTRL+C the child process running terraform is interrupted, which is undesirable for my use case.

SIGINT is now sent by default during cancellation on Unix (where SIGINT is supported).

The default timeout after cancellation is 60s after which SIGKILL is still sent. This is also adjustable as per [v0.23.0 Changelog](https://github.com/hashicorp/terraform-exec/releases/tag/v0.23.0).

Feel free to upgrade to the most recent version and let us know if this is still an issue and/or if it works for you.

PR #424, opened 2023-11-29
graph: update test output for v1.7

kmoe:


kmoe:
@radeksimko Ah yes, the eternal ambiguity between "resource" and "resource but not data source", not always helped by the rare disambiguating term "data resource"...

Issue #426, opened 2023-12-08
Consider adding show option to configure JSON decoding


bendbennett:
### Background

The [terraform-exec](https://github.com/hashicorp/terraform-exec) Go module explicitly sets the `useJSONNumber` field to true on the [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State) struct by calling [UseJSONNumber(true)](https://pkg.go.dev/github.com/hashicorp/terraform-json#State.UseJSONNumber) within the [Show()](https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.19.0/tfexec#Terraform.Show) method prior to the implicit call to [UnmarshalJSON()](https://pkg.go.dev/github.com/hashicorp/terraform-json#State.UnmarshalJSON) on [tfjson.State](https://pkg.go.dev/github.com/hashicorp/terraform-json#State). As a consequence, numerical values that are returned when the state is obtained by calling the [Show()](https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.19.0/tfexec#Terraform.Show) method on terraform-exec are of type `json.Number`.

The [tfjson.Plan](https://pkg.go.dev/github.com/hashicorp/terraform-json#Plan) struct does not currently surface a way to modify the handling of numerical values when decoding JSON, as there is no `useJSONNumber` field on the [tfjson.Plan](https://pkg.go.dev/github.com/hashicorp/terraform-json#Plan) struct, and no corresponding `UseJSONNumber()` method for setting such a field. As a consequence, numerical values that are obtained when calling methods such as [ShowPlanFile()](https://pkg.go.dev/github.com/hashicorp/terraform-exec@v0.19.0/tfexec#Terraform.ShowPlanFile) are of type `float64`.

A [PR](https://github.com/hashicorp/terraform-json/pull/113) has been opened on terraform-plugin-json which adds a `useJSONNumber` field to the [tfjson.Plan](https://pkg.go.dev/github.com/hashicorp/terraform-json#Plan) struct, adds a corresponding `UseJSONNumber()` method for setting such a field, and correspondingly modifies the `UnmarshalJSON()` method on [tfjson.Plan](https://pkg.go.dev/github.com/hashicorp/terraform-json#Plan) to take the value of `useJSONNumber` into account.

### Use-Case

[...]

PR #427, opened 2023-12-08
Add ShowOption for configuring JSON decoding

bendbennett:
Closes: #426

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/hc-install v0.6.2
-	github.com/hashicorp/terraform-json v0.18.0
+	github.com/hashicorp/terraform-json v0.18.1-0.20231207151239-da9b6fa7714c
 	github.com/zclconf/go-cty v1.14.1
 	github.com/zclconf/go-cty-debug v0.0.0-20191215020915-b22d67c1ba0b
 )
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/hc-install v0.6.2
-	github.com/hashicorp/terraform-json v0.18.1-0.20231207151239-da9b6fa7714c
+	github.com/hashicorp/terraform-json v0.19.0
 	github.com/zclconf/go-cty v1.14.1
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/google/go-cmp v0.6.0
 	github.com/hashicorp/go-version v1.6.0
 	github.com/hashicorp/hc-install v0.6.2
-	github.com/hashicorp/terraform-json v0.18.0
+	github.com/hashicorp/terraform-json v0.18.1-0.20231207151239-da9b6fa7714c
```
bendbennett, line 9: This is temporary whilst [Add useJSONNumber to Plan and amend UnmarshalJSON behavior](https://github.com/hashicorp/terraform-json/pull/113#top) is being considered.

tfexec/options.go
```diff
@@ -243,6 +243,19 @@ func GraphPlan(file string) *GraphPlanOption {
 	return &GraphPlanOption{file}
 }
 
+// JSONConfig holds information which determines how JSON is decoded.
+type JSONConfig struct {
```
kmoe, line 247: Rather than a general-purpose `JSONConfig` struct, prefer a `UseJSONNumber` struct with a single boolean option.

The design intent behind the variadic options is for each one to be explicit, thus avoiding structs whose fields may have additions over time that the caller is unaware of.

Most options are implemented this way. `ReattachConfig` below is a bit of a special case because it represents structured information that happens to be encoded in a string env var then decoded to a (hopefully) identical struct on the [other side](https://github.com/hashicorp/terraform-plugin-sdk/blob/main/plugin/debug.go#L16).

I'm not sure `UseJSONNumber` should be an option at all - it should be the default. This would be a breaking change for consumers of terraform-exec and I'm not sure yet how best to handle that.

I think a reasonable compromise in the meantime would be to make `UseJSONNumber` a single-purpose option, and then change the default to `true` as an explicit breaking change later once we're sure (v1.0 at the latest). What we need to do to be sure - I don't know.

tfexec/options.go
```diff
@@ -243,6 +243,15 @@ func GraphPlan(file string) *GraphPlanOption {
 	return &GraphPlanOption{file}
 }
 
+type UseJSONNumber struct {
```
kmoe, line 246:
```suggestion
type UseJSONNumberOption struct {
```

Naming convention for the struct.

kmoe (approved):
Approved with naming convention suggestion

go.mod
bendbennett: Bumped to `terraform-json` `v0.19.0`

PR #447, opened 2024-05-01
tfexec: Add `-allow-deferral` experimental options to `Plan` and `Apply` commands

austinvalle:
This PR introduces a new plan and apply flag `-allow-deferral`. This flag enables the deferred actions experiment that was introduced in Terraform [`v1.9.0-alpha20240404`](https://github.com/hashicorp/terraform/releases/tag/v1.9.0-alpha20240404)

### Notes
- I wasn't sure if there was an _external_ interface for checking that experiments are enabled from the Terraform CLI, so I just check for `alpha` or `dev` in the version as a best effort. Happy to update with a more accurate method if one exists!

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,7 +49,7 @@ jobs:
     name: static-checks (go ${{ matrix.go_version }})
     needs: resolve-versions
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 10
     strategy:
       fail-fast: false
       matrix:
@@ -80,7 +80,7 @@ jobs:
       - resolve-versions
       - static-checks
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 10
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,7 +49,7 @@ jobs:
     name: static-checks (go ${{ matrix.go_version }})
     needs: resolve-versions
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
     strategy:
       fail-fast: false
       matrix:
--- a/tfexec/force_unlock_test.go
+++ b/tfexec/force_unlock_test.go
@@ -5,6 +5,7 @@ package tfexec
 
 import (
 	"context"
+	"runtime"
 	"testing"
 
 	"github.com/hashicorp/terraform-exec/tfexec/internal/testutil"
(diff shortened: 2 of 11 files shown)


tfexec/force_unlock_test.go
```diff
@@ -39,6 +40,10 @@ func TestForceUnlockCmd(t *testing.T) {
 // The optional final positional [DIR] argument is available
 // until v0.15.0.
 func TestForceUnlockCmd_pre015(t *testing.T) {
+	if runtime.GOOS == "darwin" && runtime.GOARCH == "arm64" {
+		t.Skip("Terraform for darwin/arm64 is not available until v1")
+	}
+
```
austinvalle, line 46: Skipping to avoid this error: https://github.com/hashicorp/terraform-exec/actions/runs/8914271134/job/24481540878?pr=447

An unfortunate situation where the latest macos GHA runners now are macOS 14 and ARM64 architecture machines: https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/. We ran into this issue on `setup-terraform` last week: https://github.com/hashicorp/setup-terraform/pull/409

I'm not sure if we want to default fallback to `amd64` or if skipping this on `arm64` is acceptable.

.github/workflows/tests.yml
```diff
@@ -49,7 +49,7 @@ jobs:
     name: static-checks (go ${{ matrix.go_version }})
     needs: resolve-versions
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 20
```
austinvalle, line 52: Another GH workflow related change, it seems that `setup-go` is taking an extra 2 minutes to setup Go `1.22.x`, not leaving enough time for the rest of the job to run 🙂 : https://github.com/hashicorp/terraform-exec/actions/runs/8914407327/job/24482771576?pr=447

.github/workflows/tests.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - resolve-versions
       - static-checks
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 10
+    timeout-minutes: 20
```
austinvalle, line 83: Another GH workflow related change, it seems that `setup-go` is taking an extra 2 minutes to setup Go `1.22.x` on `windows-latest`, not leaving enough time for the rest of the job to run 🙂 : https://github.com/hashicorp/terraform-exec/actions/runs/8914407327/job/24482771576?pr=447

liamcervante (approved):
This looks good to me, I will chase Katy or Radek for a quick look in case I'm missing something as I haven't look closely at this before.

tfexec/version.go
```diff
@@ -180,6 +181,22 @@ func (tf *Terraform) compatible(ctx context.Context, minInclusive *version.Versi
 	return nil
 }
 
+// experimentsEnabled asserts the cached terraform version has experiments enabled in the executable,
+// and returns a well known error if not. Experiments are enabled in alpha and (potentially) dev builds of Terraform.
+func (tf *Terraform) experimentsEnabled(ctx context.Context) error {
```
radeksimko, line 186: I would've probably called this `versionIsPrerelease()` or something along these lines, as that's what the function actually does but it's also not exposed API to downstream so I'm not too worried about the name here.

tfexec/force_unlock_test.go
radeksimko: I think skipping on arm64 is acceptable given that the nature of the test is to check for version differences, not so much architecture differences, and the former is already being checked on other platforms as well.

radeksimko (approved):

PR #454, opened 2024-05-13
feat: add graceful shutdown option for apply&destroy

wjf3121:
The default behavior of CommandContext is to [immediately kill](https://cs.opensource.google/go/go/+/master:src/os/exec/exec.go;l=453;drc=1e12eab8705d1d8d7472be9147a39caa1c8380db) the command process when receiving cancellation signal from the context, which may lead to corrupted terraform state.

This PR adds the option to allow sending interruption signal to the Apply/Destroy command and wait for a graceful period before forcefully killing the process. Note that the option doesn't work on Windows due to [the lack of support](https://cs.opensource.google/go/go/+/master:src/os/exec.go;l=138-139;drc=73847aa41b9299a39e596651ccdd9c0dcb511d5d) of sending interrupt signal.


radeksimko:
Hello @wjf3121  
Thanks for the contribution and apologies for the radio silence here.

We decided to take a slightly different approach in https://github.com/hashicorp/terraform-exec/pull/512 to solve the problem which hopefully provides a simpler UX by making the graceful cancellation on by default with a default 60s timeout.

Feel free to upgrade to [v0.23.0](https://github.com/hashicorp/terraform-exec/releases/tag/v0.23.0) which was just released and contains the linked change.

PR #455, opened 2024-05-13
Reenable Dependabot for internal GitHub actions

dbanck:
When TSCCR was first introduced, it was not clear that the tool would only update external GitHub actions. This adds the configuration to enable `hashicorp/*` action updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,10 @@ updates:
     schedule:
       interval: "daily"
     labels: ["dependencies"]
+  # Dependabot only updates hashicorp GHAs, external GHAs are managed by internal tooling (tsccr)
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    allow:
+      - dependency-name: "hashicorp/*"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,14 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "daily"
+      interval: "weekly"
     allow:
       - dependency-name: "hashicorp/*"
+    groups:
+      github-actions-breaking:
+        update-types:
+          - major
+      github-actions-backward-compatible:
+        update-types:
+          - minor
+          - patch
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -5,3 +5,10 @@ updates:
     schedule:
       interval: "daily"
     labels: ["dependencies"]
+  # Dependabot only updates hashicorp GHAs, external GHAs are managed by internal tooling (tsccr)
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
xiehan, line 12:
```suggestion
      interval: "weekly"
```

IMO `daily` is unnecessarily obnoxious and it's nice if these are kept in sync with the config we use for the security-tsccr automation which is also weekly.

.github/dependabot.yml
```diff
@@ -5,3 +5,10 @@ updates:
...
+    allow:
+      - dependency-name: "hashicorp/*"
```
xiehan, line 14:
```suggestion
      - dependency-name: "hashicorp/*"
    groups:
      github-actions-breaking:
        update-types:
          - major
      github-actions-backward-compatible:
        update-types:
          - minor
          - patch
```

This is the config we use elsewhere in Core team repos which is nice because it allows you to merge the nonbreaking changes quickly while possibly taking more time to test breaking changes separately.

xiehan (commented):
Jinx, I was just about to submit the same PR. Here's where mine differs slightly from yours.

dbanck:
GitHub has recently updated the `macos-latest` runners from amd64 to arm64: https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/. Older versions of Terraform aren't available for arm64, hence failed tests.

It might make sense for this repository to use `macos-13` instead.

xiehan (approved):

bflad:
FYI, in case it helps -- hashicorp/setup-terraform v3.1.0+ will automatically install darwin/amd64 when running on darwin/arm64 for Terraform versions 1.0.2 and earlier which didn't have that release artifact. For example, we still wanted legacy Terraform version coverage with provider acceptance testing on macOS and it seemed like a pragmatic compromise until those legacy Terraform versions are officially deprecated. It'll work as long as Apple/GitHub continue to enable the CPU emulation support.

PR #457, opened 2024-05-13
github: Create CODEOWNERS file

xiehan:


radeksimko (approved):

bflad:
Are we sure this is the correct syntax?

- https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax

> If you want to match two or more code owners with the same pattern, all the code owners must be on the same line. If the code owners are not on the same line, the pattern matches only the last mentioned code owner.

Refer also to the currently working terraform-json CODEOWNERS: https://github.com/hashicorp/terraform-json/blob/5e08e15af60aeaeb1ee04686fbb2f6e9de819840/CODEOWNERS

xiehan:
Interesting, I hadn't seen that before, thanks for the callout!

PR #478, opened 2024-10-06
tfexec: add InitJSON

### LINKED ISSUE #442: Support `terraform init` JSON output (coming soon)
See https://github.com/hashicorp/terraform/pull/34886

bschaatsbergen:
Fixes #442

* refactor `tfexec/init.go` to facilitate (split some code into distinct functions) to support the new `InitJSON` function, similar to `ApplyJSON` and `PlanJSON`.
* Added `InitJSON` functions.
* Added cmd and E2E tests.


tfexec/version_test.go
```diff
@@ -300,16 +300,16 @@ func TestExperimentsEnabled(t *testing.T) {
 		tfVersion     *version.Version
 		expectedError error
 	}{
-		"experiments-enabled-in-1.9.0-alpha20240404": {
-			tfVersion: version.Must(version.NewVersion(testutil.Alpha_v1_9)),
+		fmt.Sprintf("experiments-enabled-in-%s", testutil.Latest_Alpha_v1_9): {
```
dbanck, line 303:
```suggestion
		"experiments-enabled-in-alphas": {
```

I think removing the exact version from the map key will make it more readable

tfexec/version_test.go
```diff
@@ -300,16 +300,16 @@ func TestExperimentsEnabled(t *testing.T) {
...
+			tfVersion: version.Must(version.NewVersion(testutil.Latest_Alpha_v1_9)),
 		},
-		"experiments-disabled-in-1.8.0-beta1": {
-			tfVersion:     version.Must(version.NewVersion(testutil.Beta_v1_8)),
-			expectedError: errors.New("experiments are not enabled in version 1.8.0-beta1, as it's not an alpha or dev build"),
+		fmt.Sprintf("experiments-disabled-in-%s", testutil.Latest_Beta_v1_8): {
```
dbanck, line 306:
```suggestion
		"experiments-disabled-in-betas": {
```

I think removing the exact version from the map key will make it more readable

tfexec/version_test.go
```diff
@@ -300,16 +300,16 @@ func TestExperimentsEnabled(t *testing.T) {
...
+			tfVersion:     version.Must(version.NewVersion(testutil.Latest_Beta_v1_8)),
+			expectedError: fmt.Errorf("experiments are not enabled in version %s, as it's not an alpha or dev build", testutil.Latest_Beta_v1_8),
 		},
-		"experiments-disabled-in-1.5.3": {
+		fmt.Sprintf("experiments-disabled-in-%s", testutil.Latest_v1_5): {
```
dbanck, line 310:
```suggestion
		"experiments-disabled-in-stable": {
```

dbanck (approved):
Thanks for the PR and sorry that it took us so long to review!

We're fine with the proposed API, but we expect it to change a bit in the future as we come closer to version 1.0*

radeksimko:
> We're fine with the proposed API, but we expect it to change a bit in the future as we come closer to version 1.0 *

^ it's likely there will be only one `Init()` method that by default passes the `-json` flag. We just need to figure out whether it's common enough ask for people to read the output, i.e. if we should also be unmarshaling it by default and returning some types or stream or something.

We face the same problem with `plan` and `apply` structured output to be fair - so we'll need to figure out how to model the API there as well and ideally do it the same way.

radeksimko (approved):

PR #486, opened 2024-11-21
(TFECO-8260) Add Heimdall metadata

jpogran:
This change adds the metadata for the internal Heimdall service to the repository

### DIFF
--- a/META.d/_summary.yml
+++ b/META.d/_summary.yml
@@ -1,0 +1,11 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+schema: 1.1
+partition: tf-ecosystem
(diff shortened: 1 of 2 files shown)


META.d/_summary.yml
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+schema: 1.1
+partition: tf-ecosystem
+category: library
+
+summary:
+  owner: team-tf-editor-experience
```
xiehan, line 9: From a compliance perspective, the Core team are currently considered the owners of this library. I don't mind having this be y'all, but then we should probably be consistent and have the compliance owners changed to you all as well.

PR #498, opened 2025-01-17
ci/e2etests: Add latest major Terraform versions

radeksimko:
This admittedly results in a lot of test jobs but it may be fine considering the expected number of incoming PRs.

I think it's still valuable to run tests against all major versions, we could consider doing that only on `main` if it's too much for every PR.


.github/workflows/tests.yml
```diff
@@ -140,7 +140,14 @@ jobs:
...
           - "1.2.9"
-          - "1.3.7"
+          - "1.3.10"
+          - "1.4.7"
+          - "1.5.7"
+          - "1.6.6"
+          - "1.7.5"
+          - "1.8.5"
+          - "1.9.8"
+          - "1.10.4"
```
dbanck, line 150: Would it be possible to just specify the minor version, like `1.10`, so we don't have to update it so often?

radeksimko: In theory yes but it would require some more significant refactoring in other places because right now the underlying code expects exact versions rather than constraints https://github.com/hashicorp/terraform-exec/blob/4497f9edf36dcdbbf5fc982d1d951c9509d865fd/tfexec/internal/testutil/tfcache.go#L70-L73

PR #501, opened 2025-01-20
Update CHANGELOG.md

dbanck:
I think we still should land the remaining dependency updates, especially
* https://github.com/hashicorp/terraform-exec/pull/500

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,14 +2,9 @@
 
 ENHANCEMENTS:
 * tfexec: Add support for `terraform init --json` via `InitJSON` ([#478](https://github.com/hashicorp/terraform-exec/pull/478))
-* docs: reference the plugin framework ([#480](https://github.com/hashicorp/terraform-exec/pull/480))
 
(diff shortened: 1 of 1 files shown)


dsa0x (dismissed):

dsa0x:
@dbanck But this PR only updates the CHANGELOG, right? Not the actual dependency update. Is that just preparation for applying the updates?

dbanck:
Sorry for the confusion, I meant to suggest that we should land the dependency updates before cutting a new release.

As the previous changelogs rarely listed dependency updates, I decided against including them here. If you think it's important to have them, I'll be happy to add them.

radeksimko:
I don't have strong opinion on the dependencies but would probably go with what we have done before 🤷🏻‍♂️ If you wish to set a new trend I'd also be fine with that.

dsa0x:
Totally fine with just following what we have done before. I was simply curious. Thanks @dbanck

dsa0x (approved):

PR #503, opened 2025-01-22
Update reference to Go version

mmtevelde:
After #499 was merged, I suppose this reference in the readme needs to be updated as well


dbanck (approved):
Nice catch! Thanks!

Issue #389, opened 2023-05-09
Does this library provide a way to gracefully shutdown


FLAGLORD:
If I use terraform cli and kill a running terraform task, it will gracefully shutdown.
If I use this library to invoke terraform and cancel it, will it behave the same way? If not, is there any trick?

PR #512, opened 2025-04-08
tfexec: Enable graceful (SIGINT-based) cancellation

radeksimko:
Closes https://github.com/hashicorp/terraform-exec/issues/389

## Changelog draft

- Context cancellation will now send os.Interrupt (SIGINT on Unix) instead of killing the process (which was default Go behaviour)
  - You can change the default `60s` timeout at which process is killed after attempted cancellation via `SetKillTimeout(time.Duration)`


dbanck (approved):
🚀

wjf3121:
Hi @radeksimko thanks for your effort to address this issue. However, I think this PR won't work as expected:

[`writeOutput`](https://github.com/hashicorp/terraform-exec/blob/269386fd2520f6e3ca99bb506b202d00501658cc/tfexec/cmd.go#L255) uses the context spawned from the command context, which means when the command is canceled, `writeOutput` will in turn be canceled leading to `signal: broken pipe`. This will cause the Terraform command to immediately exit without waiting for the graceful shutdown.

We found this issue when making our own implementation in #454. Hopefully it can be addressed in the main branch so we can switch to the official version again.

Another difference between this PR and #454 is that we pass `Setpgid: true` to unix cmd, which currently is only passed in the [Linux implementation](https://github.com/hashicorp/terraform-exec/blob/269386fd2520f6e3ca99bb506b202d00501658cc/tfexec/cmd_linux.go#L22). I don't recall exactly why we do this but I remember there were issues when we needed to force kill the command when graceful shutdown timed out.

radeksimko:
@wjf3121 I would be more than happy to address any outstanding issues if you can provide a way of reliably reproducing it, preferably in a form of a test.

That applies to both `signal: broken pipe` and `Setpgid: true`.

As far as I can tell, the linked PR #454 does not contain test for either of these issues so it is hard to understand the nature of those problems and confirm that the code actually solves them. Correct me if I'm wrong.

dastrobu:
I am a bit late to the party, but we see the `broken pipe` issue as well in our project. I picked up the suggestion to create a test case for this scenario. You can find it here: [https://github.com/hashicorp/terraform-exec/pull/523](https://github.com/hashicorp/terraform-exec/pull/523).

It would be great to get feedback or suggestions on how to fix the code, as the PR currently just contains the test case. Feedback is welcome.

PR #517, opened 2025-05-13
Update CODEOWNERS

austinvalle:
Our GH team name has been updated recently, so we're slowly moving over to `@hashicorp/terraform-core-plugins`

cc @bbasata

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,3 +1,3 @@
 # This codebase has shared ownership and responsibility.
 
-* @hashicorp/terraform-core @hashicorp/terraform-devex @hashicorp/tf-editor-experience-engineers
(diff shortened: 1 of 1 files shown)


radeksimko (changes requested):
I think the ACL may need to be addressed first

Screenshot 2025-05-13 at 16 01 56

.github/CODEOWNERS
```diff
@@ -1,3 +1,3 @@
 # This codebase has shared ownership and responsibility.
 
-* @hashicorp/terraform-core @hashicorp/terraform-devex @hashicorp/tf-editor-experience-engineers
+* @hashicorp/terraform-core @hashicorp/terraform-core-plugins @hashicorp/tf-editor-experience-engineers
```
austinvalle, line 3: Need to update the permissions 👍🏻

radeksimko (approved):

PR #523, opened 2025-05-21
add test to reproduce issue with signal: broken pipe on graceful shutdown

dastrobu:
## Related Issue

Relates to:

  * #454
  * #512

## Description

This is a test to show the issue with `signal: broken pipe` on graceful shutdown, introduced by #512.  
The PR only contains a test and no fix yet, as requested per [comment](https://github.com/hashicorp/terraform-exec/pull/512#issuecomment-2834540214) from @radeksimko.

The unit test `TestGracefulCancellation_interruption_on_apply` usually fails with:

`expected interrupt signal, received "signal: broken pipe"`

Although sometimes I have also seen:

`expected interrupt signal, received "exit status 255"`

It seems to depend on timing.

In our production code, we would love to see graceful shutdown being handled correctly, as intended by #512.  
However, we are observing `signal: broken pipe` and a locked Terraform state, which is what we would like to get rid of to avoid the need to manually unlock the state again.

I can try to provide a fix as well if there is some guidance on how to do it.

## Rollback Plan

  * [ ] If a change needs to be reverted, we will roll out an update to the code within 7 days.


radeksimko:
@dastrobu This looks quite promising. I'll look into this in more detail when I get a moment. Thanks a lot for sharing this test case.

PR #526, opened 2025-07-14
Run tests on all PRs

dbanck:
## Description

This PR updates the GitHub test workflow so that it runs on all PRs, not just those targeting `main`.

## Rollback Plan


dbanck:
Yup, I want tests to run in https://github.com/hashicorp/terraform-exec/pull/525

PR #527, opened 2025-07-22
arthur/fix broken pipe during shutdown

busser:
- **Update E2E test suite to test all curretn v1.x versions of Terraform**
- **Add E2E test that shows graceful shutdown is broken**
- **Fix graceful shutdown**


radeksimko:
@busser Thank you for the deep investigation and a very helpful and detailed PR description, much appreciated. 🙏🏻 

I ran the whole matrix of tests which raised some failures unfortunately. Could you look into these failures?

--- 

While we focus on fixing those I would be happy to accept & merge 7434949b92d6e757aaecb857b6045907ed88e10d in a separate PR if you can raise it. Feel free to cc me on that PR and I'll try to merge it in a timely manner so you can rebase here and keep this PR focused on solving the bug. Regarding the tests run time - we can always trim some versions if it becomes a problem but I think the trade-offs here are in favour of correctness over speed considering the current cadence of contributions to the project.

radeksimko:
I'm wondering if these failures could be simply a result of a flakey test as it is dependent on `time.Sleep`?

If that is the case, we could try wrapping it in the new experimental [`testing/synctest`](https://pkg.go.dev/testing/synctest@go1.25rc2) potentially, unless we find some other way of removing the flakiness.

As to how we could introduce it - Go 1.25 is around the corner (expected for August, i.e. in a few weeks). We could wait until the next Go release, bump the version and just use it. Alternatively I'd also be open to modifying the CI workflows such that we can use it as an experiment.

busser:
Thanks for looking into this! I investigated the test failures and I don't think it's a flaky test issue. Rather, these failures are related to something I mentioned earlier: Terraform pre-1.1 doesn't seem to react when sent an interrupt signal.

The tests fail specifically for Terraform versions 0.12 to 1.0 inclusive. The `TestContext_sleepTimeoutExpired` test (the only one that fails) is already skipped for versions under 0.12, and starting with version 1.1, Terraform shuts down correctly when signaled.

I don't think Terraform ignoring the signal has anything to do with `tfexec` itself. The issue doesn't discriminate based on OS, so it's likely not a platform-specific signaling issue. The test fails now because before this PR's fix, Terraform crashed due to the broken pipes, which the test considered a successful shutdown. Now that the crash is gone, tests detect that Terraform doesn't shut down properly for these older versions.

To address this, I'll add a commit to skip that test for versions up to 1.0, similar to the test I introduced in this PR.

I'll also open a separate PR with the commit that adds all the versions to test as you suggested, so we can get that merged first and keep this PR focused on the bug fix.

radeksimko:
@busser The 1.1 behaviour change is a fantastic find. 👏🏻 

I managed to find specifically the PR which introduced the change: https://github.com/hashicorp/terraform/pull/29825 - as verified by rebuilding Terraform with and without that commit and re-running your test which fails without that patch.

That PR was actually missed from the public changelog, perhaps because it was difficult to tell how _exactly_ it impacted end users and what they can possibly do about it. This is now - sic retrospectively - a bit clearer.

--- 

Whilst cancellations <1.1 were not graceful, the library would still honour the cancellation in some shape or form, i.e. calling code of tfexec would **not** be left hanging.

In this PR, the `TestContext_sleepTimeoutExpired` fails after 10 seconds w/ Terraform <1.1, i.e. it would hit the 10s (or any longer) timeout in the test. I think that implies that consumers of tfexec using it alongside Terraform <1.1 would now be left hanging on attempted cancellation, which does not seem ideal. 🤔 

Having looked into why it is left hanging, I could observe a zombie process which is left hanging even after the `Kill()` initiated via `WaitDelay`. 😑 

Of course closing the pipes early (before cancellation) like we do now (prior to this PR) is still wrong and we need to fix that.

I just can't think of any decent backwards-compatible solution right now. Need to think about this a bit more. Constructive ideas welcomed.

busser:
@radeksimko Ready for your review :)

radeksimko (approved):
LGTM, Thank you for working on this and for talking through all the details at length.

Much appreciated.

PR #531, opened 2025-08-18
(Re)Implement `QueryJSON` as `iter.Seq[NextMessage]`

radeksimko:
## Example Usage

```go
iter, err := tf.QueryJSON(context.Background())
if err != nil {
	t.Fatalf("error running Query: %s", err)
}

for nextMsg := range iter {
	if nextMsg.Err != nil {
		// handle error
		break
	}
	switch m := nextMsg.Msg.(type) {
	case tfjson.ListStartMessage:
		// do something
	case tfjson.ListResourceFoundMessage:
		// do something
	case tfjson.ListCompleteMessage:
		// do something
	}
}
```

PR #538, opened 2025-09-12
Handle terraform command and stdout errors

dsa0x:
## Related Issue

Fixes # 

## Description

In plain English, describe your approach to addressing the issue linked above. For example, if you made a particular design decision, let us know why you chose this path instead of another solution.

## Rollback Plan


radeksimko:
This was incorporated into https://github.com/hashicorp/terraform-exec/pull/531 via cherry-pick

Thanks for the suggestion

PR #539, opened 2025-09-16
Implement `QueryJSON` and introduce new way for consuming Terraform's structured logging

dbanck:
## Description

This PR adds a new internal method that allows easier consumption of Terraform's structured log format. It exposes log messages as an iterator via `iter.Seq[NextMessage]`.

`QueryJSON` is the first command that makes use of new the method.

## Rollback Plan


SarahFrench (approved):
:shipit:

PR #590, opened 2026-05-14
feat: adds noColor support instead of always passing `-no-color` by default

ramonvermeulen:
## Related Issue
Fixes (partly): https://github.com/magosproject/magos/issues/50

Later PR in https://github.com/magosproject/magos will follow with the magos side of the implementation.


ramonvermeulen:
My bad, wrong repo.

PR #610, opened 2026-09-22
Add opt-in hiding of Terraform command windows on Windows

Shubham-Padkonde:
## Related Issue

Fixes #570.

## Description

Add opt-in `Terraform.SetHideWindow(true)` for Windows GUI applications embedding Terraform. Command construction sets Windows `SysProcAttr.HideWindow` when enabled. The default is unchanged, the setting can be disabled again, and it has no effect on non-Windows platforms.

The option follows the existing instance-level setter API and applies to all commands, including internal version detection. It does not expose arbitrary process attributes or change cancellation behavior.

Validation on Windows, Go 1.27.1:
- Regression for default/enabled/disabled behavior fails before wiring the option into command construction and passes afterward.
- Hidden child-process execution test passes.
- A real Terraform 1.14.0 `Version` call with the option enabled passes.
- Existing command helper tests selected alongside the new tests pass.
- `go vet ./...`, formatting and `git diff --check` pass. Linux test binary cross-compiles; it was not executed on Linux.
- The broader unit run fails on existing tests: missing standalone `echo`, a pre-0.15 force-unlock expectation, and a nil command in the 0.13 graph test. All three reproduce with the original command builder. Full unit/e2e/race success is not claimed.

Prepared with OpenAI Codex assistance.

## Rollback Plan

Release timing is controlled by maintainers; this additive option can be reverted without changing the default process behavior.

None. This controls Windows command-window visibility only; environment, logging, credentials, and command permissions are unchanged.


crw:
Thanks for your interest. It is unlikely this will be reviewed at this time, I would suggest proposing your change on the issue and see if there is maintainer interest in reviewing a contributed solution. Thanks!

PR #611, opened 2026-09-22
Add workspace select OrCreate option

### LINKED ISSUE #575: Support "-or-create" in WorkspaceSelect
Request to add support for WorkspaceSelect option that enables "-or-create" flag, allowing for more efficient idempotent workspace management.

Shubham-Padkonde:
Fixes #575.

Expose `tfexec.OrCreate(true)` for `WorkspaceSelect`, so automation can select an existing workspace or create a missing one with Terraform's native `-or-create` behavior. The default and explicit false preserve the original command, including compatibility with older Terraform. Enabling the option returns a wrapped ErrVersionMismatch before invoking workspace select on Terraform < 1.4.

Validation:
- All workspace command unit tests pass, including enabled, disabled, minimum supported version, and unsupported-version cases.
- All workspace end-to-end tests pass against local Terraform 1.14.0 on Windows. The new case verifies that false does not create a missing workspace, true creates it, and true selects it again after switching away.
- `go vet ./...`, gofmt, and `git diff --check` pass.

The full unrelated unit/e2e matrix was not rerun; this change was tested with the workspace suites. No real cloud resources are created. Prepared with Codex assistance.


crw:
Thanks for your interest. It is unlikely this will be reviewed at this time, I would suggest proposing your change on the issue and see if there is maintainer interest in reviewing a contributed solution. Thanks!

